PR #106, opened 2022-04-08
Fix compilation and tests

aloisklink:
Fixes issues with:

  - sqlite static linking (we need to manually specify that `-lm` (math.h), `-lpthread`, and `-ldl` is required)
  - pcap static linking (we need to manually specify that `-lnl-genl-3 -lnl-3` is required)
    - I've disabled PCAP DBUS and RDMA support, since otherwise PCAP only used DBUS and RDMA if they're installed,
      which means static linking changes. Let me know if DBUS/RDMA are required. Tests seem to run fine without them though.
    - The pkgconfig file for pcap says that we also need to link with `libssl` and `libcrypto` if using pcap statically, but it seems to work without me specifying anything so :shrug:
  - fixes the test_engine test_init_context as `config_ifinfo_array` was added

I've tested with the following, and all tests pass on my machine.

(side-note, building on my desktop instead of a laptop is sooooooooo much faster, make sure to modify `"-DLIB_MAKEFLAGS=--jobs=8"` to however many cores you want to use)

```bash
mkdir build/ && cd build/
# a bunch of these defines probably aren't needed anymore
cmake .. -DCMAKE_LIBRARY_ARCHITECTURE='x86_amd64'  -DCROSS_COMPILE_EDGESEC=OFF  -DBUILD_MNL_LIB=ON -DBUILD_NETLINK_LIB=ON -DBUILD_NL_LIB=ON -DUSE_UCI_SERVICE=OFF -DUSE_GRPC_LIB=ON -DBUILD_SQLSYNC_SERVICE=ON -DBUILD_REVERSE_SERVICE=ON -DUSE_PTHREAD_LIB=ON -DUSE_M_LIB=ON -DBUILD_HOSTAPD=ON -DBUILD_REST_SERVER=ON -DBUILD_MICROHTTPD_LIB=ON "-DLIB_MAKEFLAGS=--jobs=8"  -DUSE_NETLINK_SERVICE=ON
make -j4
make test
```


src/engine.c
```diff
@@ -278,6 +278,11 @@ int init_context(struct app_config *app_config, struct supervisor_context *ctx)
   }
 
   utarray_new(ctx->config_ifinfo_array, &config_ifinfo_icd);
+  if (app_config->config_ifinfo_array == NULL) {
+    errno = EINVAL;
+    log_err("Invalid app config, missing config_ifinfo_array");
+    return -1;
```
aloisklink, line 284: For most of your error handling you use `log_trace()` to print errors.

I felt like `log_err()` made more sense for an error condition (especially since `log_trace()` isn't enabled by default unless you change the logging levels), but you're welcome to change this to `log_trace()` instead if you feel like this is better.

src/engine.c
```diff
@@ -278,6 +278,11 @@ int init_context(struct app_config *app_config, struct supervisor_context *ctx)
   }
 
   utarray_new(ctx->config_ifinfo_array, &config_ifinfo_icd);
+  if (app_config->config_ifinfo_array == NULL) {
```
aloisklink, line 281: Potentially a dumb `C` question, but can `app_config->config_ifinfo_array` be uninitialized, and therefore be filled with random data??

Not sure how to check that it's not a valid `utarray` without a segmentation fault, though.

src/engine.c
mereacre: Thanks. I will have to change to log_trace, because log_err uses errno.

mereacre (approved):

src/engine.c
mereacre: `app_config->config_ifinfo_array` can be freed and then it is counted as uninitialized.

`utarray` library does check. On allocation error it exist the process https://github.com/nqminds/EDGESec/blob/main/src/utils/utarray.h#L108

lib/pcap.cmake
```diff
@@ -35,17 +39,20 @@ if (BUILD_PCAP_LIB AND NOT (BUILD_ONLY_DOCS))
     find_library(LIBPCAP_LIB NAMES libpcap.a pcap PATHS "${LIBPCAP_LIB_DIR}" NO_DEFAULT_PATH)
   endif ()
 
-  # static pcap needs -libverbs -lnl-genl-3 -lnl-3  -ldbus-1
+  # static pcap needs -lnl-genl-3 -lnl-3 # maybe openssl and crypto too???
```
mereacre, line 42: As far as I know, pcap shouldn't depend on these libraries.

mereacre (approved):

src/engine.c
aloisklink: Isn't `log_err` appropriate in this situation? I set the `errno` to [EINVAL, aka "Invalid argument"](https://en.cppreference.com/w/cpp/error/errno_macros#:~:text=EINVAL)

`log_trace()` is the lowest level of logging, below even `log_debug()`, so it only appears if you do `-dddddd`.

mereacre: Yes, one can use it like that. But a created `log_err` to be used only for system error that come out of glib library commands or kernel call. For engine.c `log_debug` would be the right one to use.

src/engine.c
aloisklink: Correct me if I'm wrong, (last time I programmed in C was ages ago), but if I do the following:

### `app_config.config_ifinfo_array` is unitialized (junk data)

Maybe undefined behaviour?? I'm not sure if C can catch this issue.

```c
app_config app_config;
app_config.param1 = 1;
init_context(&app_config, ...)
```

### `app_config.config_ifinfo_array` is initalized to 0 (NULL ptr)

Causes a segmentation fault at the following line (which is why I added the `if()` check)

https://github.com/nqminds/EDGESec/blob/99bedafa65c5519207408b79504df4cf1f59d3fa/src/engine.c#L281

```c
app_config app_config = {0, .param1 = 1};
init_context(&app_config, ...)
```

### `app_config.config_ifinfo_array` is correctly initalized to a utarray

Expected valid behaviour by `init_context`

```c
app_config app_config = {0, .param1 = 1};
utarray_new(app_config.config_ifinfo_array, &ut_str_icd);
init_context(&app_config, ...)
```

## Potential cleanup

Instead of having a heap-allocated utarray using `UT_array*` and using `utarray_new` or `utarray_free`, maybe it's better to have a stack-allocated `UT_array` (e.g. not a pointer), and using `utarray_init` and `utarray_done` instead? 

https://github.com/nqminds/EDGESec/blob/56299399c6f8d8eed97397ea675e496160189ec6/src/config.h#L47

lib/pcap.cmake
aloisklink: After you removed these lines in https://github.com/nqminds/EDGESec/pull/106/commits/99bedafa65c5519207408b79504df4cf1f59d3fa, compiling still worked on my PC.

However, my `./build/lib/pcap/lib/pkgconfig/libpcap.pc` file says `Libs.private:  -lnl-3 -l/usr/lib/x86_64-linux-gnu/libssl.so -l/usr/lib/x86_64-linux-gnu/libcrypto.so`, so it should need `lnl-3` `libssl` and `libcrypto` for static linking.

Looking at `libpcap`'s build file, we can see that if we're compiling `libpcap` for Linux, it will automatically build netlink support, see https://github.com/the-tcpdump-group/libpcap/blob/master/CMakeLists.txt#L1331-L1348

Maybe PCAP does require linking to `nl-3`, but we're already linking it somewhere?? Not sure.

aloisklink: If you're sure that PCAP will never need `netlink` support, I'll compile `libpcap` with `BUILD_WITH_LIBNL=OFF` so that we never have to worry `netlink` support.

mereacre: Lets compile it with `BUILD_WITH_LIBNL=OFF` for now.

PR #112, opened 2022-04-11
style: Remove EOL whitespace

aloisklink:
Most editors like VS Code automatically delete trailing whitespace when you save a file.

`git diff --ignore-space-at-eol` confirms that trailing EOL whitespace are the only ones that have been removed.

I've made this as a separate PR, since I've noticed in my other PRs, there's loads of these changes.

@mereacre, do you think it's worth setting up something like `clang-format` to format the C and C++ code? See https://clang.llvm.org/docs/ClangFormat.html

It's what a lot of big C/C++ code-bases do. It can automatically style/lint our code to use a specific style: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

(see https://zed0.co.uk/clang-format-configurator/ for a quick and easy way to see what styles there are, my personal preferences are LLVM or Google style, but you're the C programmer here so you should pick the style you like)

### DIFF
--- a/src/ap/ap_config.h
+++ b/src/ap/ap_config.h
@@ -18,11 +18,11 @@
  ****************************************************************************/
 
 /**
- * @file ap_config.h 
- * @author Alexandru Mereacre 
+ * @file ap_config.h
+ * @author Alexandru Mereacre
  * @brief File containing the definition of AP config structures.
- * 
- * Defines the access point (AP) configuration structure used to configure the 
+ *
+ * Defines the access point (AP) configuration structure used to configure the
  * AP service.
  */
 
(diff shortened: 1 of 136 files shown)


mereacre:
Let's go for the LLVM style. Could you please set it up?

mcr:
You can also setup git to refuse to do commits when there is trailing whitespace.
I recommend doing that as well if feasible.

aloisklink:
> Let's go for the LLVM style. Could you please set it up?

Awesome, I'll set it up in another PR.

> You can also setup git to refuse to do commits when there is trailing whitespace.
> I recommend doing that as well if feasible.

That's a good idea too. Although there might be a conflict with `docs/docusaurus`, as it adds git pre-commit hooks to automatically style markdown. `pre-commit` has a hook for `clang-format`, so it would make it easy to format code automatically too: https://github.com/pre-commit/mirrors-clang-format

PR #125, opened 2022-04-26
Fix glibc version failure in cmake

aloisklink:
On my computer, with gcc-11 (default) there is no glibc filename that contains the version number of glibc.

This causes `cmake` to fail on my computer.

It's possible to easily find out the glibc version by compiling and running a program, however, this won't work without emulation when cross-compiling.

Because of this, instead we use GNU strings to search for all strings in the glibc library.

The library lists all the versions it supports in the format GLIBC_x.y.z, so all we have to do is some simple string opts to find the latest version, then to return that version.


CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ include(glibc)
 
 if (NOT CROSS_COMPILE_EDGESEC)
   CHECK_GLIBC_VERSION()
+  IF (GLIBC_VERSION STREQUAL "2.2.1")
+    MESSAGE(FATAL_ERROR "Unsupported version of glibc: ${GLIBC_VERSION}")
+  ENDIF ()
```
aloisklink, line 31: Would this be a better error message? (I'm guessing we want older versions to also fail)

Also, is it worth adding a comment why glibc <= 2.2.1 doesn't work?

```suggestion
  IF (GLIBC_VERSION VERSION_LESS_EQUAL "2.2.1")
    MESSAGE(FATAL_ERROR "Unsupported version of glibc: ${GLIBC_VERSION}")
  ENDIF ()
```

mereacre: That make sense.

aloisklink: Huh, it's weird you encountered this issue. [glibc version 2.2.1 is apparently from 2001](https://sourceware.org/legacy-ml/libc-alpha/2001-01/msg00163.html), which is before OpenWRT, Ubuntu, or git even existed.

mcr:
Why do you need to know the glibc version?
It seems just failing to compile would be fine, if it's too old.

aloisklink:
> Why do you need to know the glibc version?
> It seems just failing to compile would be fine, if it's too old.

It's always better to have a human readable error message when something goes wrong.

However, I'm also a bit confused as to why the version check is actually needed. The GLIBC version we're checking and throwing an error for is [glibc version 2.2.1, which came out in 2001](https://sourceware.org/legacy-ml/libc-alpha/2001-01/msg00163.html), so it's nothing something you should encounter, unless you're trying to compile on a Windows 98 device.

@mereacre, do you remember why you added this check? My gut feeling is it might be a typo?

I've recently added some lines to CMake that will make it throw an error unless C11 can be used, so we can probably get rid of these `glibc` checks, as [C11 support was only added in glibc 2.16](https://sourceware.org/legacy-ml/libc-alpha/2012-06/msg00807.html) https://github.com/nqminds/EDGESec/blob/f5da46af9acc4617d020725fe4761b92356e4bec/CMakeLists.txt#L144-L145

mereacre:
@Alois let's remove the check. I might have added it a long time ago as I had a problem with compiling a library.

PR #126, opened 2022-04-26
Add `cmake-preset` for Linux

aloisklink:
Currently, manually defining all the different EDGESec defines for each platform is quite complicated.

The new [cmake-presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) feature in new versions of CMake makes this a lot easier, by allowing us to pre-define some presets.

Essentially, as long as you have CMake v3.22.0 or greater, you can do the following to configure/build/test easily:

```bash
cmake --preset linux # configure EDGESec for Linux in a ./build/linux folder
cmake --build --preset linux -j4 # build EDGESec for Linux using 4 threads
ctest --preset linux # test EDGESec for Linux
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,11 @@ repos:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
+    -   id: check-json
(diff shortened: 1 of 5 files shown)


mereacre (approved):
Very cool feature. Thanks ;)

PR #127, opened 2022-04-26
Add a cmake-preset for `openwrt`

aloisklink:
Creates a CMakeToolchain file that automatically downloads the OpenWRT 19.07.10 mvebu/cortexa9 SDK, and uses that for cross-compilation.

Additionally, I've added a cmake-preset for it, and it seems to work just with the following, but I haven't tested it yet:

```bash
# old behaviour is available at `cmake --preset openwrt/default`
cmake --preset openwrt/mvebu/cortexa9
cmake --build --preset openwrt # builds in ./build/openwrt/mvebu/cortexa9
```

I haven't bothered running `tests` yet, but if you're interested in using this on your computer, @mereacre, you can try and see if the changes actually work.

~I haven't yet tested whether the built `edgesec` binary actually runs on OpenWRT!! Especially since the SDK is from vanilla OpenWRT; it might not work on Turris's fork!!.~ Seems like the OpenWRT 19.07.10 SDK works to compile for the Turris Omnia fork, see  https://github.com/nqminds/EDGESec/pull/127#issuecomment-1116454292


mcr:
OpenWRT does not use glibc.  Are we building with glibc?
At present 90% of Turris has been upstreamed to OpenWRT, so they share most of the fundamental pieces, but Turris has a different package manager and image build process.  This matters if one wants to get things packaged into the image, for instance because one wants to generate an identity on first boot.

aloisklink:
> OpenWRT does not use glibc. Are we building with glibc?

Not for the OpenWRT build using the `CMakeModules/CMakeToolchains/openwrt-mvebu-cortexa9.cmake` file I just added.
That uses `muslgnueabi`, which is what comes with the OpenWRT SDK for mvebu/cortexa9 19.07.10
https://github.com/nqminds/EDGESec/blob/222ddfa3bfec9931122084ab98eb03fbfb4222aa/CMakeModules/CMakeToolchains/openwrt-mvebu-cortexa9.cmake#L28-L30

After having a look at https://gitlab.nic.cz/turris/os/build, I'm pretty sure this should work with Turris OS on the Turris Omnia, but I'll need to test it on Monday to confirm.

The `glibc` check in #125 was only active when compiling locally for testing (i.e. on Ubuntu). But I'm about to remove it since it seems redundant.

> At present 90% of Turris has been upstreamed to OpenWRT, so they share most of the fundamental pieces, but Turris has a different package manager and image build process. This matters if one wants to get things packaged into the image, for instance because one wants to generate an identity on first boot.

:+1: The way we're compiling right now is a bit hacky, we're compiling and statically linking `libubox`/`libuci` from the latest `master`. Ideally we should use the OpenWRT/Turris OS packaging system, so that we can instead use dynamic linking on the versions of `libubox`/`libuci` actually on OpenWRT 19.07.x.

aloisklink:
> I would prefer a docker container for cross compiling and other staff

100%. Docker is useful since you still need to `sudo apt install` a bunch of dependencies. CMake still recommends putting cross-compiling settings into their own toolchain file, even if we use a Docker container.

> Can we remove/comment for now the mvebu/cortexa9 SDK part?

In commit https://github.com/nqminds/EDGESec/pull/127/commits/f67dd7f1baa8ae35e4b5887353f6a9af2be4e939, I've changed the cmake-preset so it's now called `openwrt/mvebu/cortexa9`, and I've added the old config as just `openwrt` (where the cross-compiler is installed under `/usr/local`)

So:
  - `cmake --preset openwrt` is the same behaviour as the old `mkdir build && cd build && cmake .. && make`, where it uses `/usr/local/bin/arm-linux-musleabihf-*`, and has an error if it's not installed.
  - `cmake --preset openwrt/mvebu/cortexa9` is the one that downloads the OpenWRT SDK automatically

If you let me know where you got your `/usr/local/bin/arm-linux-musleabihf-*` toolchain from, I can test it locally and make a `Dockerfile`.

mereacre:
I made a container for the toolchain here: https://github.com/nqminds/EDGESec/tree/main/deployment/containers

aloisklink:
Hey @mereacre:

I compiled the `openwrt/mvebu/cortexa9` code, and then copied over the `edgesec` and `capsrv` executables to the Turris Omnia OpenWRT router:

```bash
scp build/openwrt/mvebu/cortexa9/config.ini build/openwrt/mvebu/cortexa9/src/edgesec build/openwrt/mvebu/cortexa9/src/capsrv turris:/tmp/edgesec_openwrt-cmake-prefix-6ff921c62
```

Then I ran the new `edgesec` code compiled with the OpenWRT SDK, but using the old `config.ini`, and it outputs the same results as the `edgesec_v0.9.9_alpha5_arm32` exe, so it looks like it works!

```console
root@turris:/tmp/edgesec_openwrt-cmake-prefix-6ff921c62# ./edgesec -c ../edgesec_v0.9.9_alpha5_arm32/config.ini  -s 12345 -dddddddd
[...]
2022-05-03 11:30:44.598  ERROR[ENOTCONN(107)] domain.c:208: [Socket not connected] sendto
2022-05-03 11:30:44.599  TRACE supervisor.c:347: SET_IP fail
2022-05-03 11:30:45.118  TRACE supervisor.c:336: Supervisor received 59 bytes from socket length=0
2022-05-03 11:30:45.118  TRACE cmd_processor.c:336: Received DHCP request with type=arp
2022-05-03 11:30:45.118  TRACE cmd_processor.c:344: Wrong type
2022-05-03 11:30:45.118  TRACE domain.c:205: Sending to socket on 
```

I've also updated the `deployment/containers/README.md` instructions to use the new `openwrt/default` preset.
All I had to do to get it working is change the Dockerfile to use `ubuntu:22.04`, since that has a newer version of CMake.

mereacre (approved):

PR #135, opened 2022-05-23
build(ath79/generic): support ath79/generic

aloisklink:
Add support for compiling for ath79/generic on OpenWRT 19.07.10

Try it out with `cmake --preset openwrt/ath79/generic` and `cmake --build -j4 --preset openwrt/ath79/generic`

The https://openwrt.org/toh/hwdata/tp-link/tp-link_eap225-outdoor_v1 is not currently supported, as it requires at least OpenWRT 21.02.

I'm currently trying to download the OpenWRT 21.02 SDK, but the office internet is slow, so it's going to take a while.

### DIFF
--- a/CMakeModules/CMakeToolchains/openwrt-21-ath79-generic.cmake
+++ b/CMakeModules/CMakeToolchains/openwrt-21-ath79-generic.cmake
@@ -1,0 +1,56 @@
+cmake_minimum_required(VERSION 3.9.0)
+
+set(CMAKE_SYSTEM_NAME               Linux) # OpenWRT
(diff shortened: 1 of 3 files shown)


aloisklink:
I've added support for compiling for ath79/generic on OpenWRT 21.02.03

Try it out with `cmake --preset openwrt-21/ath79/generic`.

It looks like the EDGESec and capsrv binaries are both 1 MiB larger, so EDGESec is now about 7 MiB! I think it's probably compiled in Debug mode, not Release mode, so we should be able to reduce the size.

mereacre (approved):
Great

PR #136, opened 2022-05-23
Compile `libnetlink` at compile time

aloisklink:
Previously, we compiled libnetlink using a cmake child process during the configure step.

Now, we instead compile libnetlink during the cmake build process using the original cmake executable.

This will make compiling faster, as the `cmake` configure step is single-processed, while `cmake --build` can use multiple processes.

To make this easier, I've set it up so that `libnetlink` is always compiled as a `STATIC` library, as it's only ever used by the `src/utils/nl.c` file.

This simplifies a lot of the code.


mereacre (approved):
Great.

PR #137, opened 2022-05-23
Compile `sqlite3` in compile step

aloisklink:
Use `ExternalProject_Add()` to compile SQLite3, if needed.

`ExternalProject_Add()` automatically downloads, configures, compiles, and runs `make install` during the `cmake --build` step of the parent project.

Other stuff:
  - Throws an error if BUILD_SQLITE_LIB=OFF, and SQLite3 cannot be found.

    BREAKING CHANGE: The minimum `cmake` version to compile EDGESec
      is now cmake v3.14.0, instead of v3.13.0, as it is the first version that adds support
      for FindSQLite3.cmake. This is a 3 year old version that came out in March 2019, so I don't think this should matter much.
  - Prevents installing sqlite3 if it's a static lib


src/supervisor/CMakeLists.txt
```diff
@@ -9,51 +9,47 @@ add_library(mac_mapper mac_mapper.c)
...
-  target_link_libraries(cmd_processor PRIVATE system_commands crypt_commands monitor_commands network_commands sqlite_fingerprint_writer mac_mapper domain net log os firewall_service iface base64)
+  target_link_libraries(cmd_processor PUBLIC supervisor_config PRIVATE system_commands crypt_commands monitor_commands network_commands sqlite_fingerprint_writer mac_mapper domain net log os firewall_service iface base64)
 else ()
-  target_link_libraries(cmd_processor PRIVATE system_commands monitor_commands network_commands sqlite_fingerprint_writer mac_mapper domain net log os firewall_service iface base64)
+  target_link_libraries(cmd_processor PUBLIC supervisor_config PRIVATE system_commands monitor_commands network_commands sqlite_fingerprint_writer mac_mapper domain net log os firewall_service iface base64)
 endif ()
 
+add_library(supervisor_config INTERFACE)
+set_target_properties(supervisor_config PROPERTIES PUBLIC_HEADER "supervisor_config.h")
+target_link_libraries(supervisor_config INTERFACE SQLite::SQLite3)
```
aloisklink, line 51: I added this as an "Interface" aka header-only library, since many files depend on `supervisor_config.h`, but there is no `supervisor_config.c`.

That way, if we ever remove "sqlite.h` from this library (or add another depedency), we only need to change it in one place.

src/CMakeLists.txt
```diff
@@ -13,42 +13,40 @@ add_subdirectory(capture)
 
 add_library(sqlhook SHARED sqlhook.c)
 target_compile_options(sqlhook PRIVATE -fPIC)
-target_include_directories(sqlhook PRIVATE ${LIBSQLITE_INCLUDE_DIR})
-target_link_libraries(sqlhook PRIVATE domain os SQLite::sqlite)
+target_link_libraries(sqlhook PRIVATE domain os SQLite::SQLite3)
```
aloisklink, line 16: I've removed all of the `target_include_directories(... LIBSQLITE_INCLUDE_DIR)`, since they get automatically added when we use `target_link_libraries(SQLite::SQLite3)`

mereacre (approved):
Great work.

PR #138, opened 2022-05-24
build(install): fix install when disabled crypto

aloisklink:
When running `cmake --install` with -DUSE_CRYPTO_SERVICE=OFF,
Linux tries to install `/`, aka the entire root directory.

This is because `${LIBOPENSSL_LIB_PATH}/` will open to `/`, as `LIBOPENSSL_LIB_PATH` is undefined.


mereacre (approved):
Great

PR #140, opened 2022-05-24
cmake `libuuid` dependency improvement

aloisklink:
Two changes:
  - First, `uuid` is now built at compile time using `ExternalProject`.
    This allows `make -j4` to build it with multiprocessing, and makes
    `cmake` a lot faster, since the hard work is now during `cmake --build`.
  - Secondly, `-DBUILD_UUID_LIB=OFF` now works.
    I copied `CMakeModules/FindLibUUID.cmake` from https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/Modules/FindLibUUID.cmake
    to search for LibUUID. It's a BSD 3-Clause license, so all we need to use it is include the license/copyright info in our binary (already added to the `debian/copyright` file)


mereacre (approved):
Great work.

PR #141, opened 2022-05-24
Compile `pcap` in compile step

aloisklink:
We use `FetchContent_MakeAvailable(libpcap)` to compile `libpcap` at compile time.

Additionally, we only compile `pcap_static`, and nothing else.

Since this is a static library, we also don't need to include it in `cmake --install`.


mereacre (approved):
Great

PR #144, opened 2022-05-27
cmake `libuci` dependency improvement

aloisklink:
Two changes:
  - First, `uci` and `ubox` is now built at compile time using `ExternalProject`.
    This allows `make -j4` to build it with multiprocessing, and makes
    `cmake` a lot faster, since the hard work is now during `cmake --build`.
  - Secondly, `-DBUILD_UCI_LIB=OFF` now works.
    This will make UCI/UBOX compiliation optional, and should hopefully mean that EDGESec now compiles correctly with the OpenWRT SDK.
    - [ ] Todo, currently testing this with the OpenWRT SDK


mereacre (approved):
Great work

PR #147, opened 2022-05-31
Compile `openssl` at compile time

aloisklink:
Use `ExternalProject_Add()` to compile OpenSSL::Crypto at compile time, if `-DWITH_CRYPTO_SERVICE=ON` and `-DBUILD_OPENSSL_LIB=ON`. This lets all of the `downloads`/`make` of all the ExternalProjects run at the same time.

Additionally, instead of using `find_library()` to look for OpenSSL::Crypto when `-DBUILD_OPENSSL_LIB=OFF`, instead I'm using the built-in [FindOpenSSL](https://cmake.org/cmake/help/v3.14/module/FindOpenSSL.html), which allows us to specify that we must have at least version OpenSSL 3.0.0.

I've also added `-DWITH_CRYPTO_SERVICE=ON` and `-DBUILD_OPENSSL_LIB=ON` to `--preset linux`, so that we're still testing the crypto code.


mereacre (approved):
Great

PR #152, opened 2022-06-10
Various CMake cleanups

aloisklink:
Various CMake cleanups:

- [build: depend cmake options on parent options](https://github.com/nqminds/EDGESec/commit/14360e5718f59da8508476fbb988e706e7009cb3)
  Some CMake options only make sense to enable/show to the user if another CMake option is enabled.
  
  For example, setting `BUILD_MNL_LIB=ON`/`OFF` only makes sense if we're actually using libmnl when `USE_NETLINK_SERVICE=ON`
- [build: remove LIB_MAKEFLAGS references](https://github.com/nqminds/EDGESec/commit/aba44f8c162ea892f41e25ffad9cf327803fc0ab)
  `LIB_MAKEFLAGS` was used to pass makeflags to libraries that were being compiled at configure time.
  
  As of https://github.com/nqminds/EDGESec/pull/150 being merged, all libraries are now compiled at compile time, so this variable is no longer needed.
- [build: remove unused CMake defines](https://github.com/nqminds/EDGESec/commit/a80509e33d0edb55f31880ab652e2600d6ba0149)
  `CROSS_COMPILE_EDGESEC` had been replaced by `CMAKE_CROSS_COMPILING` and `BUILD_NL_LIB` is unused.
- [build(libnetlink): allow setting CMAKE_BUILD_TYPE](https://github.com/nqminds/EDGESec/commit/dd916f90d78fbfaccf9f096176658d8594304f8c)
  Previously, libnetlink was hard coded to compile in DEBUG mode.
  Now it should inherit EDGESec's compile mode.


mereacre (approved):
Great

PR #155, opened 2022-06-13
build: fix cmake library definitions

aloisklink:
Adds missing CMake library definitions.

I've added a CMake definition for some header-only libraries too, like ap_config.h and capture_config.h, as since they require utils/os.h, they need to link to the os.c lib.


mereacre (approved):
Super

mereacre (approved):

PR #157, opened 2022-06-13
Fix `cmake --preset openwrt/mvebu/cortexa9`

### LINKED ISSUE #156: `cmake --preset openwrt/mvebu/cortexa9` no longer works
Running `cmake --preset openwrt/mvebu/cortexa9` causes an error:

```
CMake Error at CMakeLists.txt:3 (project):
  The CMAKE_C_COMPILER:

    /home/alois/Documents/EDGESec/build/openwrt/mvebu/cortexa9/_deps/openwrt_sdk_mvebu_cortexa9-src/staging_dir/toolchain-arm_cortex-a9+vfpv3-d16_gcc-7.5.0_musl_eabi/bin/arm-openwrt-linux-muslgnueabi-gcc

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
```

Last known working commit: c527013cb7a4ceb438adb2ba4e1a4f54a7a0bb79
Earliest known broken commit: 78ed0dcadc80891ffb008d8354ecd5ac86ae3e4a

```bash
git bisect start 78ed0dcadc80891ffb008d8354ecd5ac86ae3e4a c527013cb7a4ceb438adb2ba4e1a4f54a7a0bb79
# deletes build folder, except for `*.tar.*`-ish, then tries configure
git bisect run sh -c "find ./build/ -type f -name '*.tar.*' -delete; /snap/bin/cmake --preset openwrt/mvebu/cortexa9"
git bisect reset
```

EDIT: It looks like the issue is due to https://github.com/nqminds/EDGESec/commit/49e30002194262651cfb8448a7c13f0fd1b31613

For some reason, the `ExternalProject_Add` `PREFIX` option works fine only the first time you run `cmake` (hence why GitHub Actions is happy to run this code).
The second time you run `cmake`, something breaks.

aloisklink:
Fixes #156.

In the FetchContent call for the OpenWRT SDK, I set `PREFIX` to a fixed folder relative to the toolchain file:

https://github.com/nqminds/EDGESec/blob/5c6ab26d15ef02288b9940b2ec098bdc19f731f3/CMakeModules/CMakeToolchains/openwrt-mvebu-cortexa9.cmake#L19-L21

This was to prevent CMake from re-downloading the OpenWRT SDK into a temporary directory, when running a CMake sub-process (e.g. used internally by CMake in `CMakeDetermineCompilerABI.cmake` to detect the C/C++ ABI).

However, for some reason (maybe a bug in CMake?), this `PREFIX` code worked the first time you ran `cmake` (hence with GitHub Actions was succeeding) but failed when running it for a second time.

I've changed the code to now instead always download to the `build/` directory.
Instead, I'm declaring an environment variable that's used by the CMake sub-processes to avoid re-downloading the OpenWRT SDK.


mereacre (approved):
Looks good.

mereacre:
It works. The problem was the build folder has root permission.

The only error that I get is this one:
```bash
[100%] Built target edgesec
CMake Error: No such test preset in /home/alex/Projects/EDGESec: "openwrt/mvebu/cortexa9"
Available test presets:

  "default"
  "linux"
```
But the binary is compiled.

PR #161, opened 2022-06-15
Fix process sign blob cmd typo

aloisklink:
Minor typo fix for the compile errors on the `main` branch, see errors https://github.com/nqminds/EDGESec/runs/6902926349?check_suite_focus=true

It looks like there are also tests that are broken, as sending `OK\n` now adds a newline and sends `OK\n\n`, but that can always be fixed later :)


aloisklink:
Ah, whoops, I made this PR too late, the build issues have already been fixed in 0a9c4cdcea33113a4a773a990456ee1fb15aa09b

PR #162, opened 2022-06-15
Fix failing `cmd_processor` tests

aloisklink:
A lot of the mocked `cmd_processor` functions were mocked to return `OK_REPLY`.
Then, the tests were checking to make sure that exactly `strlen(OK_REPLY)` bytes were written.

However, https://github.com/nqminds/EDGESec/commit/260584216ba1eefd0bb9d7c2ca6c9e4dbc5a7a35 added a newline to many writes, so instead we need to check for `strlen(RESULT "\n")`.

### DIFF
--- a/src/supervisor/cmd_processor.h
+++ b/src/supervisor/cmd_processor.h
@@ -313,7 +313,9 @@ ssize_t process_query_fingerprint_cmd(int sock,
  * @param client_addr The client address for replies
  * @param context The supervisor structure instance
  * @param cmd_arr The array of received commands
- * @return ssize_t Size of reply written data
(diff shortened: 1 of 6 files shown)


mereacre (approved):
Very good. One comment. You could use os_zalloc instead of malloc and adding a null charcater at the end.

aloisklink:
> You could use os_zalloc instead of malloc and adding a null charcater at the end.

It seems like it would be faster to use `malloc` and only add a null character at the end, instead of adding null characters everywhere and replacing it later.

But, knowing C, `malloc` and `os_zalloc` might be the same speed :)

PR #163, opened 2022-06-17
refactor: tidy up close_capture_thread

aloisklink:
Fixes a C compiler warning in `close_capture_thread`.

For some reason, the C compiler thought that `current->value.capture_pid` was a pointer type,
and was printing a warning on my PC about comparing a pointer type using `> to and int `0`.

I'm guessing maybe `HASH_ITER` was causing some weird behaviour with the C compiler?

Maybe `pthread_t` has a pointer type in OpenWRT or something, since the docs say that the structure is purposely never described :shrug: 

> ```c
> /* Thread identifiers.  The structure of the attribute type is not
>    exposed on purpose.  */
> typedef unsigned long int pthread_t;
> ```
> From my local `/usr/include/x86_64-linux-gnu/bits/pthreadtypes.h`

I've changed the loop (since `HASH_ITER` should only be used if you're deleting data, see https://troydhanson.github.io/uthash/userguide.html#_iterating_and_sorting), made the input `const`, and replace the `> 0` check with a `if (pid == 0) continue` check, and that seems to have fixed the warning.

It might also be worth renaming the variable from `capture_pid` to `capture_thread`.
The documentation from [`<sys/types.h>`](https://pubs.opengroup.org/onlinepubs/7908799/xsh/systypes.h.html) shows that PIDs are described with `pid_t`, not `pthread_t`, so calling it a PID might be confusing.

### DIFF
--- a/src/runctl.c
+++ b/src/runctl.c
@@ -369,14 +369,14 @@ int run_mdns_forwarder(char *mdns_bin_path, char *config_ini_path) {
   return ret;
 }
 
-void close_capture_thread(hmap_vlan_conn **vlan_mapper) {
-  hmap_vlan_conn *current, *tmp;
-
-  HASH_ITER(hh, *vlan_mapper, current, tmp) {
-    if (current->value.capture_pid > 0) {
-      if (pthread_join(current->value.capture_pid, NULL) != 0) {
-        log_errno("pthread_join");
-      }
+void close_capture_thread(const hmap_vlan_conn *vlan_mapper) {
+  for (const hmap_vlan_conn *current = vlan_mapper; current != NULL;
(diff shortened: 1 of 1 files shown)


mereacre (approved):
The docs say that HASH_ITER is a safe loop if one wants to remove elements. So, it cold be as well as used to loop. But I like your version. It is neater.

mereacre:
Agree with the pid defs. I will change to tid.

aloisklink:
> The docs say that HASH_ITER is a safe loop if one wants to remove elements.

That's true. Maybe it's the C++ guy in me, but pre-processor macros are very scary to me :)

image

PR #164, opened 2022-06-17
refactor: make `sign_blob_cmd` args `const`

aloisklink:
The data pointed to by `keyid` and `blob` are read-only, so we should declare it as `const`.

We can always undo it if we need to, but C will warn us first.

Added while I was debugging https://github.com/nqminds/EDGESec/pull/162


mereacre (approved):
Great.

PR #168, opened 2022-06-20
fix: create folders for SQLite macconn_db

aloisklink:
Running edgesec on a real Turris Omnia hub failed, due to the folders to macconn_db not existing.

Because of this, this PR automatically builds the folders to `macconn_db`, if they do not exist.


mereacre:
Let's keep it open and I will add to it. Because we need to do the same for crypt and capure db. We will also need to add an option in the config.ini to create folders to path.

aloisklink:
Bump! I just encountered this issue when testing the edgesec `.deb`

```
2022-08-02 09:43:48.963  DEBUG runctl.c:323: Opening the macconn db=/var/lib/edgesec/connection.sqlite
2022-08-02 09:43:48.963  DEBUG sqlite_macconn_writer.c:38: Cannot open database: unable to open database file
2022-08-02 09:43:48.963  ERROR runctl.c:326: open_sqlite_macconn_db fail
```

mereacre (approved):

PR #169, opened 2022-06-20
ci: add build/test preset for crypto service

aloisklink:
Currently, none of our CI attempts to build or run the crypto service.
This adds a preset for it, so that it is tested correctly.


mereacre (approved):
Great

PR #172, opened 2022-06-24
Add CMake function for custom capture middlewares

### LINKED ISSUE #165: Add CMake macro for custom capture middlewares
Currently, middlewares are hard-coded into the capture service.

https://github.com/nqminds/EDGESec/tree/main/src/capture

Ideally, we'll want a cmake macro that automatically adds middlewares.

E.g. something like:

```cmake
# declares library pcap_middleware
include("./pcap_middleware")
edgesec_add_middleware(pcap_middleware)
```

aloisklink:
Closes #165

Adds a CMake function called `edgesecAddCaptureMiddleware()` that allows users of EDGESec to create their own capture middlewares.

Users just need to declare a CMake library that contains a `struct capture_middleware` that contains function pointers to the `init()`, `process()`, and `free()` functions for their middleware.

Then, they can use the following CMake function for EDGESec to automatically link their middleware to the capture service.

```cmake
   edgesecAddCaptureMiddleware(
        MIDDLEWARE_TARGET <CMAKE_TARGET_CONTAINING_YOUR_MIDDLEWARE>
        MIDDLEWARE_STRUCT <C_NAME_OF_MIDDLEWARE_VARIABLE>
   )
```

Additionally, I've added a CI action to test building with all of the example capture middlewares.

### Example Usage

Example `my_middlewares.h`

```c
#include "middleware.h"
extern struct capture_middleware my_first_middleware;
extern struct capture_middleware my_second_middleware;
```

Example: `my_middleware.c`

```c
#include "my_middleware.h"
struct capture_middleware my_first_middleware = {
  .init = my_middleware_init_function,
  .process = my_middleware_process_function,
  .free = my_middleware_cleaner_function,
};
struct capture_middleware my_second_middleware = {
  // ...
};
```

Example: `CMakeLists.txt`
```cmake
add_library(my_middlewares src/my_middlewares.c)
include(EDGESec)

edgesecAddCaptureMiddleware(
    MIDDLEWARE_TARGET my_middlewares
    MIDDLEWARE_STRUCT my_first_middleware
)
edgesecAddCaptureMiddleware(
    MIDDLEWARE_TARGET my_middlewares
    MIDDLEWARE_STRUCT my_second_middleware
)
```

### Questions

I've added documentation for this in file for `src/capture/CMakeLists.txt`, but that's a bit hard to find. Is there a better place to document this? Maybe we should make a new docs on the website called `ADVANCED_USAGE.md` and place this here.


src/capture/cleaner_middleware/cleaner_middleware.h
```diff
@@ -26,47 +26,13 @@
...
- * @param context The middleware context
- * @param ltype The packet type
- * @param header The pcap packet header
- * @param packet The pcap packet
- * @param ifname The capture interface
- * @return int 0 on success, -1 on failure
+ * @brief Cleaner Middleware.
+ * @author Alexandru Mereacre, Alois Klink
+ * The cleaner middleware is designed to periodically remove the oldest
+ * PCAP files when the use more than `CLEANER_STORE_SIZE` KiB.
```
aloisklink, line 35: I think this is what the cleaner middleware does, but you might want to double-check this description.

src/capture/header_middleware/header_middleware.h
```diff
@@ -26,50 +26,11 @@
...
- * @brief Header packet processors
- *
- * @param context The middleware context
- * @param ltype The packet type
- * @param header The pcap packet header
- * @param packet The pcap packet
- * @param ifname The capture interface
- * @return int 0 on success, -1 on failure
+ * @brief Header Middleware.
+ * @author Alexandru Mereacre, Alois Klink
```
aloisklink, line 33: We should add a description as to what the header middleware does. I'm guessing it stores the packet header data in the capture.sqlite database?

src/capture/pcap_middleware/pcap_middleware.h
```diff
@@ -26,49 +26,11 @@
...
- * @brief Pcap processor
- *
- * @param context The middleware context
- * @param ltype The packet type
- * @param header The pcap packet header
- * @param packet The pcap packet
- * @param ifname The capture interface
- * @return int 0 on success, -1 on failure
+ * @brief PCAP Middleware.
+ * @author Alexandru Mereacre, Alois Klink
```
aloisklink, line 33: Same here, we should add a description of what the PCAP middleware does.

src/capture/header_middleware/header_middleware.h
mereacre: Exactly. It stores the packet headers accordingly to the given schema.

src/capture/pcap_middleware/pcap_middleware.h
mereacre: It stores the raw packet (header + data) in files with metadata in the capture.sqlite

mereacre (approved):
Very nice.

mereacre:
A couple of remarks:
 - `middleware.h` and `middlewares.h` they almost look the same and can become cofusing. We could rename `middlewares.h{.c.in}` to `middlewares_list.h{.c.in}` or similar.
 - Perhaps we should create subfolder called `middlewares` and move all `*_middleware` subfolders to it including `middleware.h` and `middlewares.h`. Also `middlewares` we should create another CMakeLists.txt and add all the middleware creatioon, etc function. In this way we will separate the capture service from the middleware service.

aloisklink:
> * `middleware.h` and `middlewares.h` they almost look the same and can become cofusing. We could rename `middlewares.h{.c.in}` to `middlewares_list.h{.c.in}` or similar.

Done in 80b740fa4a3ffc6d1375ad44d2f8c083b29bdbd5

> * Perhaps we should create subfolder called `middlewares` and move all `*_middleware` subfolders to it including `middleware.h` and `middlewares.h`. Also `middlewares` we should create another CMakeLists.txt and add all the middleware creatioon, etc function. In this way we will separate the capture service from the middleware service.

I've made a subfolder called `src/capture/middlewares` and moved the `*_middleware` folders into it in 2b1dd11aa1b68e59d06a3c3b69979890c04fd93e.

However, I've let the `middleware.h` and `middlewares_list.h` (+ middleware creation code) in `src/capture`. I feel like `middlewares` should only be used for optional implementations/examples of middlewares.

Plus, if people want to write their own custom middlewares, they'd need to include the `middleware.h` file, my gut feeling is that the easiest way to include it will be with:

```c
#include <edgesec/capture/middleware.h>
// NOT #include <edgesec/capture/middlewares/middleware.h>
```

(side note, it looks like `src/dns/mdns_mapper.h` and `src/dns/mdns_service.c`  both require files from `src/capture/middlewares/header_middleware/mdns_decoder.h`, which seems kinda weird)

#### Other changes

- In 3fd6c4eb10ebef8fbf6d083f7688c96b529ba1d8, I added some doxygen docs to `middlewares_list.h`, but doxygen doesn't build HTML docs for `static` functions by default, so you can't see some of the docs.
- In d9b51469970a1b95ecde5f0c9de7a9bf07dc0921, I fixed some VS Code settings. Now the [CMake tools plugin](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) is used to find the paths to `#include`, which will stop the Cannot find `sqlite3.h` vs code warnings.

src/capture/header_middleware/header_middleware.h
aloisklink: Documented in 703faf635ad3cbe868087f843f082bc942129e0f

src/capture/pcap_middleware/pcap_middleware.h
aloisklink: Documented in 703faf635ad3cbe868087f843f082bc942129e0f
I also added that is it recommended to run the `cleaner_middleware` too, as the PCAP data stores a lot of bytes.

mereacre (approved):
Great

Issue #177, opened 2022-06-27
Licence and tool rename


mereacre:
Rename every EDGESec to edgesec.

PR #179, opened 2022-06-27
Compress license header in files

aloisklink:
Instead of having a massive license header in our source files, I compressed them into tags supported by Doxygen. This means that the copyright info can now be seen in Doxygen.

The new copyright info is:

```c++
/**
 * @copyright
 * SPDX-FileCopyrightText: © {$YEAR} NQMCyber Ltd and edgesec contributors
 * SPDX-License-Identifier: LGPL-3.0-or-later
 */
```

The SPDX ID is what's used by many open-source projects, such as the Linux kernel: https://lwn.net/Articles/739183/

Additionally, I've removed the `<filename>` from `@file <filename>`, since Doxygen defaults to the current filename, is the `<filename>` specifier is empty.

Does part of #177 (the license renaming part)

### DIFF
--- a/src/ap/ap_config.h
+++ b/src/ap/ap_config.h
@@ -1,21 +1,11 @@
-/****************************************************************************
- * Copyright (C) 2020 by NQMCyber Ltd                                       *
- *                                                                          *
- * This file is part of EDGESec.                                            *
- *                                                                          *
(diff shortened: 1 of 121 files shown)


mereacre (approved):
Great

PR #180, opened 2022-06-28
Rename CMake project from `EDGESEC` to `edgesec`

aloisklink:
Working on #177

Btw, it looks like the standard practice for naming projects in CMake is CamelCase, e.g. https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_VERSION.html#variable:CMAKE_PROJECT_VERSION

Is it worth us naming the CMake project `Edgesec` instead? (I'll be honest, it looks very weird to me, but it might be worth it just to make it consistent with the rest of the CMake world)


mereacre:
Camel case does look a bit weird. Let's keep it to edgesec.

PR #181, opened 2022-06-28
Use CMake's `project()` function

aloisklink:
CMake automatically parses the `VERSION` arg of `project()` into `PROJECT_VERSION_MAJOR`, `PROJECT_VERSION_MINOR`, `PROJECT_VERSION_PATCH`

See https://cmake.org/cmake/help/latest/command/project.html


mereacre (approved):
Can we generated badges automatically based on version number?

mereacre:
That's a good plan. Let's delete the previous versions.

PR #182, opened 2022-07-01
build: only build static or shared sqlite lib

aloisklink:
Only build the static or the shared version of sqlite.

Building sqlite is one of the slowest parts of the build process. We can cut down on the time the sqlite build takes by only building the shared library version or the static library versioning, depending on which version we need.

Please be aware, that due to multiprocessing, this will only cut down speed on low processor count computers, e.g. CI servers. Desktops/laptops probably won't be affected.


mereacre:
Do we need to set the default value for BUILD_SHARED_LIB? Maybe set it to OFF in openwrt preset.

aloisklink:
https://github.com/nqminds/edgesec/blob/3064b421ec18614aae48420f0fb864b76809a455/CMakeLists.txt#L118-L119

It's off by default for edgesec, since `BUILD_STATIC_LIB` fails due to cyclical linking dependencies that GCC can't handle.

PR #183, opened 2022-07-01
Remove CA.pem file and `edgesec-common` debian output

aloisklink:
The CA.pem file was used for the old packages that are no longer included in the edgesec repository. @mereacre, can you confirm that the `CA.pem` file is no longer needed?

The `/etc/edgesec/config.ini` file is now included in the `edgesec` package, as all other executable/binary packages that reply on `/etc/edgesec/config.ini` are gone. This means we can delete the `edgesec-common` package.


mereacre:
CA no longer nedded.

mereacre (approved):
Great

PR #184, opened 2022-07-01
Refactor test_mac_mapper.c initialiser

aloisklink:
GCC was warning about unbalanced struct initalisers in `test_mac_mapper.c`.

I've updated the initialiser to use designator expressions: (see https://en.cppreference.com/w/c/language/struct_initialization),
to make it a bit more clear what's going on.

In doing so, I think I encountered a bug in the tests.
The initialiser `{'b', 'r', '0', '\0'}` was setting `ip_sec_addr`.

I think this was is supposed to be setting `ifname` instead.
It's just that the tests weren't updated since https://github.com/nqminds/edgesec/commit/10baca5baf97ee3d6eb1b9ae3d198caff718f2a6, which added a new field before .ifname.

### DIFF
--- a/tests/supervisor/test_mac_mapper.c
+++ b/tests/supervisor/test_mac_mapper.c
@@ -18,60 +18,52 @@
 
 static const UT_icd mac_conn_icd = {sizeof(struct mac_conn), NULL, NULL, NULL};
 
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Thanks for sorting out the tests.

aloisklink:
No worries :)

GCC was spitting out a lot of big warnings, so I had to fix it to make it quiet!

PR #185, opened 2022-07-01
Remove `libedgesec` deb package

aloisklink:
Removes the `libedgesec` deb package.

The libedgesec debian package is no longer required.

All of edgesec dependencies are either in the `Depends:` section, or are statically linked to the edgesec binary.

The only shared library we still need is `libmnl`, however it's available as a `Depends:` dependency, so we don't need to compile/package it oursevles.


mereacre (approved):
Great

PR #191, opened 2022-07-04
build(docs): bump DOT_GRAPH_MAX_NODES to 128

aloisklink:
Currently, there are 84 nodes in the include diagram for src/utils/os.h and we need to manually increase the limit if we want it to render.

After increasing the limit, we can see this very ~spaghetti-like~ beautiful diagram by looking at https://nqminds.github.io/edgesec/os_8h.html 
image

I mainly increased the limit to hide the warning that Doxygen printed.


mereacre:
Impressive diagram.

mereacre (approved):

PR #196, opened 2022-07-08
Fix some more doxygen errors

aloisklink:
Fixes some more doxygen errors.

I was bored waiting for stuff to compile, so while I was waiting, I fixed the rest of the Doxygen errors.

It looks like all the remaining Doxygen errors are from `eloop.h` or `src/radius/*`, which use a different non-Doxygen documentation system ([kernel-doc](https://docs.kernel.org/doc-guide/kernel-doc.html)).


mereacre (approved):
Great

PR #200, opened 2022-07-20
Sqllock

mereacre:
Solves the problem of sqlite race, where multiple threads are trying to create a table on the same database.

Solution:
 - Add a busy timeout, where sqlite pools for a given time (milliseconds)
 - Change `CRATE TABLE` to `CREATE IF NOT EXISTS`

### DIFF
--- a/src/capture/middlewares/header_middleware/sqlite_header.c
+++ b/src/capture/middlewares/header_middleware/sqlite_header.c
@@ -605,146 +605,58 @@ int init_sqlite_header_db(sqlite3 *db) {
 
   log_debug("sqlite autocommit mode=%d", sqlite3_get_autocommit(db));
 
-  rc = check_table_exists(db, "eth");
-
-  if (rc == 0) {
-    log_debug("eth table doesn't exist creating...");
-    if (execute_sqlite_query(db, ETH_CREATE_TABLE) < 0) {
-      log_error("execute_sqlite_query fail");
-      return -1;
-    }
-  } else if (rc < 0) {
-    log_error("check_table_exists fail");
+  if (execute_sqlite_query(db, ETH_CREATE_TABLE) < 0) {
+    log_error("execute_sqlite_query fail");
     return -1;
   }
 
-  rc = check_table_exists(db, "arp");
-
-  if (rc == 0) {
-    log_debug("arp table doesn't exist creating...");
-    if (execute_sqlite_query(db, ARP_CREATE_TABLE) < 0) {
-      log_error("execute_sqlite_query fail");
-      return -1;
-    }
-  } else if (rc < 0) {
-    log_error("check_table_exists fail");
+  if (execute_sqlite_query(db, ARP_CREATE_TABLE) < 0) {
+    log_error("execute_sqlite_query fail");
     return -1;
   }
 
-  rc = check_table_exists(db, "ip4");
-
-  if (rc == 0) {
-    log_debug("ip4 table doesn't exist creating...");
-    if (execute_sqlite_query(db, IP4_CREATE_TABLE) < 0) {
-      log_error("execute_sqlite_query fail");
-      return -1;
-    }
-  } else if (rc < 0) {
-    log_error("check_table_exists fail");
+  if (execute_sqlite_query(db, IP4_CREATE_TABLE) < 0) {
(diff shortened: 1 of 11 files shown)


src/capture/middlewares/header_middleware/sqlite_header.c
```diff
@@ -595,156 +595,65 @@ int save_packet_statement(sqlite3 *db, struct tuple_packet *tp) {
...
-  if (rc == 0) {
-    log_debug("eth table doesn't exist creating...");
-    if (execute_sqlite_query(db, ETH_CREATE_TABLE) < 0) {
-      log_error("execute_sqlite_query fail");
-      return -1;
-    }
-  } else if (rc < 0) {
-    log_error("check_table_exists fail");
+  if (execute_sqlite_query(db, ETH_CREATE_TABLE) < 0) {
+    log_error("execute_sqlite_query fail");
```
aloisklink, line 606: Looks good!

It might be worth replacing most of this function with:

```c
  const char* create_table_commands[] = {
    ETH_CREATE_TABLE,
    ARP_CREATE_TABLE,
    IP4_CREATE_TABLE,
    IP6_CREATE_TABLE,
    TCP_CREATE_TABLE,
    UDP_CREATE_TABLE,
    ICMP4_CREATE_TABLE,
    ICMP6_CREATE_TABLE,
    DNS_CREATE_TABLE,
    MDNS_CREATE_TABLE,
    DHCP_CREATE_TABLE,
  };

#ifndef ARRAY_SIZE
#define ARRAY_SIZE(X) sizeof((X))/sizeof((X)[0])
#endif

  for (int i = 0; i < ARRAY_SIZE(create_table_commands); i++) {
    if (execute_sqlite_query(db, create_table_commands[i]) < 0) {
      log_error("execute_sqlite_query fail: %s", create_table_commands[i]);
      return -1;
    }
  }
```

It would definitely make it easier to add/remove more tables (but the error message would need to say exactly which table failed)

src/utils/sqliteu.c
```diff
@@ -14,33 +14,43 @@
...
+    log_error("Failed to execute statement: %s", sqlite3_errmsg(db));
     return -1;
   }
 
   return 0;
 }
 
-int check_table_exists(sqlite3 *db, char *table_name) {
-  sqlite3_stmt *res;
+int prepare_find_table(sqlite3 *db, char *table_name, sqlite3_stmt *res) {
```
aloisklink, line 30: We're not editing this string, so we might as well make it `const`.

```suggestion
int prepare_find_table(sqlite3 *db, const char *table_name, sqlite3_stmt *res) {
```

src/capture/middlewares/header_middleware/sqlite_header.c
mereacre: Agree. Less code is better. Smaller chances for something to fail.

Thanks for the suggestion. I will change it.

src/capture/middlewares/header_middleware/sqlite_header.c
```diff
@@ -595,66 +595,25 @@ int save_packet_statement(sqlite3 *db, struct tuple_packet *tp) {
...
-
-  if (execute_sqlite_query(db, MDNS_CREATE_TABLE) < 0) {
-    log_error("execute_sqlite_query fail");
-    return -1;
-  }
-
-  if (execute_sqlite_query(db, DHCP_CREATE_TABLE) < 0) {
-    log_error("execute_sqlite_query fail");
-    return -1;
+  for (size_t i = 0; i < STRLEN(tables); i++) {
```
aloisklink, line 612: Will `STRLEN` here work? After all, tables is an array of pointers.

aloisklink: Also, good idea in using `size_t` instead of `int` (since `int` might be only 16-bits on some embedded devices)

mereacre: https://github.com/nqminds/edgesec/blob/main/src/utils/os.h#L45

aloisklink: Ah, good catch :+1:

:facepalm: of course, a string is just an array of chars

PR #201, opened 2022-07-22
Docker

mereacre:
Moved the contenst of docker to root folder. Updated README.md


README.md
```diff
@@ -86,6 +86,66 @@ To enable verbose debug mode use:
...
+
+```console
+alias docker=podman
+```
+
+### Configure
+
+#### Step 1
+
+In `config.mak` change the `TARGET` key to the desired architecture (see details [https://github.com/richfelker/musl-cross-make](https://github.com/richfelker/musl-cross-make)):
```
aloisklink, line 101: Tbh, I'd rather there be a `--build-arg TARGET=...` in the `docker build` command, (see [`docker build --build-arg`](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)) but that can be another PR/issue.

README.md
```diff
@@ -86,6 +86,66 @@ To enable verbose debug mode use:
 ./build/src/edgesec -c ./build/dev-config.ini -ddddd
 ```
 
+## Docker for cross compile
+
+The following commands also work using the `podman` container runtime (recommended on Linux):
+
+```console
+alias docker=podman
+```
```
aloisklink, line 95: `console` should only be used if you have a syntax like the following (e.g. with PS1, command, and stdout/stderr):

```console
user@server:/folder-path/to/dir$ echo "hello"
hello
```

If it's just the command, you normally want `bash` or `sh`

`````suggestion
```bash
alias docker=podman
```
`````

aloisklink (approved):
Looks good to me!

I'll be honest, I'm not a super big fan of the `Dockerfile`, but this PR doesn't change anything, it just moves it, so there's no point in changing anything now.

README.md
mereacre: Yes, lets make another PR/issue.

PR #207, opened 2022-07-26
Added changelog update for 0.0.8

mereacre:
Added changelog update for 0.0.8


aloisklink:
Before merging, can you also update the `debian/changelog` file?

It might also be worth deleting all the old `0.9.9-...` entries too.

That way instead of having version `1:0.0.8` we can use `0.0.8` directly (without the `1` in front of it)

aloisklink (approved):
Looks good!

Just before you merge, please run `git commit --amend` and add a commit message for the `debian/changelog` message.

If we publish v0.0.8, it will point to that commit, so it needs to have a good message.

Maybe something like `chore: release v0.0.8 debian`

PR #208, opened 2022-07-28
ci(docs): use new github-pages CI action

aloisklink:
GitHub Pages now accepts being deployed directly from GitHub Actions, so we no longer need a `gh-pages` branch.

See https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/


mereacre (approved):
Super cool.

aloisklink:
Hi @cbrafter, this is already merged, but you may still want to review this, just so you're aware on how we can upgrade old repos that are still using `gh-pages`.

Issue #210, opened 2022-07-28
edgesec binary produced for TARGET=aarch64-linux-musl fails to run


katarapkokid:
**Describe the bug**

edgesec binary produced for TARGET=aarch64-linux-musl fails to run on target platform  OpenWrt 21.02.1, r16325-88151b8303 on RPi 3B+

This is related to build process referenced in issue #209 

**To Reproduce**
Steps to reproduce the behavior follow from those listed in #209 

1. With the correct compiler information build the system:

```
# docker run --rm --volume "$PWD":/opt/EDGESec --workdir /opt/EDGESec openwrt cmake --build --preset openwrt/default -j4
```

This produces -- 
```
user@linux /home/user/edgesec/build/openwrt/default/src# ls -ltr
total 7208
-rw-r--r-- 1 root root   12310 Jul 28 12:03 Makefile
-rw-r--r-- 1 root root     408 Jul 28 12:03 CTestTestfile.cmake
-rw-r--r-- 1 root root    2623 Jul 28 12:03 cmake_install.cmake
drwxr-xr-x 7 root root    4096 Jul 28 12:03 CMakeFiles
drwxr-xr-x 3 root root    4096 Jul 28 12:05 dhcp
drwxr-xr-x 3 root root    4096 Jul 28 12:05 firewall
drwxr-xr-x 3 root root    4096 Jul 28 12:05 radius
drwxr-xr-x 3 root root    4096 Jul 28 12:05 utils
-rwxr-xr-x 1 root root  156256 Jul 28 12:05 libsqlhook.so
drwxr-xr-x 3 root root    4096 Jul 28 12:05 ap
-rw-r--r-- 1 root root   34906 Jul 28 12:05 libconfig.a
drwxr-xr-x 4 root root    4096 Jul 28 12:05 capture
drwxr-xr-x 3 root root    4096 Jul 28 12:05 supervisor
drwxr-xr-x 3 root root    4096 Jul 28 12:05 dns
-rw-r--r-- 1 root root   25092 Jul 28 12:05 librunctl.a
-rwxr-xr-x 1 root root 7033360 Jul 28 12:05 edgesec

```

2. copy edgesec to targert system (scp)

3. Run it
```
root@OpenWrt:~/ManySecured# ./edgesec
./edgesec: line 1: syntax error: unexpected ")"
```

5. Check ldd
```
root@OpenWrt:~/ManySecured# ldd ./edgesec 
[...]

PR #211, opened 2022-07-28
OpenWRT Raspberry Pi 3 Toolchain Preset

aloisklink:
Adds the CMakePreset `openwrt-21.02.1/bcm27xx/bcm2710`, which is compiled for `bcm27xx/bcm2710` (Raspberry Pi 3) for `openwrt-21.02.1`, as discussed in https://github.com/nqminds/edgesec/issues/210#issuecomment-1198112852

Not currently documented (other than by running `cmake --list-presets`)

Ideally we would document how to create new presets from OpenWRT SDK releases, instead of us trying to individually add every single preset ourselves.


mereacre (approved):
Looks great.

PR #212, opened 2022-07-28
Use `debian/control` to install dependencies

aloisklink:
Our `debian/control` file has all the files required to build edgesec on a Debian PC. We should use this to control all dependency management.


mereacre (approved):
Do we still docker?

aloisklink:
> Do we still docker?

I don't think we officially support it, but I think we should set one up, mainly for people on Windows/MacOS.

I can update it today.

PR #214, opened 2022-07-29
build(openwrt): fix broken tests for OpenWRT

aloisklink:
Make skips compiling tests when cross-compiling, so you will only encounter this error when compiling for x86_64 OpenWRT.

The issue was that `test_runctl.c` could not find "uci.h".

Adding `supervisor_config` as a dependency to `test_runctl` adds UCI to the include paths when needed.

(I also fixed the `dnsmasq` bit, since that had a linking error too)


mereacre (approved):
Great

PR #216, opened 2022-07-29
chore(deps): bump hostapd to v2.10

aloisklink:
Bump hostapd to v2.10 instead of just using a git commit.

Support was added for OpenSSL 3.0 in hostapd v2.10, so we could try doing something with that too.

Full release notes are at:
https://lists.infradead.org/pipermail/hostap/2022-January/040148.html

### DIFF
--- a/lib/hostapd.cmake
+++ b/lib/hostapd.cmake
@@ -28,14 +28,12 @@ if (BUILD_HOSTAPD AND NOT (BUILD_ONLY_DOCS))
 
   ExternalProject_Add(
     hostapd_externalproject
-    # can't use https since debian build doesn't like the https certificats of https://w1.fi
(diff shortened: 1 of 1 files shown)


mereacre (approved):

aloisklink:
In my opinion this is pretty low risk, and we can just merge this straight away without testing it (especially since we only use our own compiled hostapd on Debian)

aloisklink:
I'm going to merge this, testing edgesec on Ubuntu 22.04 is failing since only `libssl3` is available, there is no `libssl1.1`

Currently, the only part of the `edgesec` that uses `libssl` is `hostapd` (unless you enable `USE_CRYPTO_SERVICE=true`).

And `hostapd` v2.10 conveniently adds support for OpenSSL 3 :)

PR #225, opened 2022-08-02
Add code coverage GitHub Action

aloisklink:
It seems like the `CMake/CodeCoverage.cmake` file wasn't being used, so I quickly added a GitHub Action for it.

Currently, the coverage action runs after all tests pass, since it only makes sense to check coverage if tests pass. I've also kept it as a separate `cmake` action, as it requires `lcov`, which is not normally installed by people.

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -61,6 +61,35 @@ jobs:
           retention-days: 7
           path: |
             ${{ runner.temp }}/edgesec-${{ matrix.cmake-preset }}/
+  coverage:
+    name: Code Coverage
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # we need to be able to write to a PR to make a comment
+      pull-requests: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install Dependencies
+        shell: bash # we're using bash arrays here
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/create-debs.yml
```diff
@@ -61,6 +61,35 @@ jobs:
...
+          sudo mk-build-deps --install --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes'  debian/control
+      - run: cmake --preset coverage
+      - run: cmake --build --preset coverage
+      - name: Report code coverage
+        uses: zgosalvez/github-actions-report-lcov@v1
+        with:
+          coverage-files: build/coverage/*.info
+          minimum-coverage: 0
+          artifact-name: code-coverage-report
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
aloisklink, line 91: By the way, if you want to use https://about.codecov.io/ (might as well, since it's free for open-source repos), all we'll have to do is sign up for it, then modify the code coverage action to look like this.

```suggestion
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v3
        with:
          files: build/coverage/coverage.info
          fail_ci_if_error: true
```

mereacre: Great.

mereacre (approved):
Great ideas.

PR #227, opened 2022-08-03
fix: fix missing WITH_NETLINK_SERVICE in iface

aloisklink:
It looks like the WITH_NETLINK_SERVICE wasn't being defined properly, so iface was compiling with nothing enabled. This was causing the compiler to emit a warning about `params` not being used.

I've added an `#error` check too if not iface service is enabled, to confirm the behaviour.


mereacre (approved):
Looks good.

PR #229, opened 2022-08-03
Run code coverage on more devices/presets

aloisklink:
This PR changes the CI code to automatically calculate code coverage on all tested code.

Then, another CI job is run to combine the code coverage for the multiple presets into a single coverage file, which is then uploaded.

This means that if we do start testing OpenWRT code in GitHub actions, it will be automatically uploaded too.

It's also faster, since instead of having to build once for tests, and once for coverage, we can do both in one step.

**Currently a draft, since this PR depends on #226**


mereacre (approved):
Great

PR #230, opened 2022-08-03
Fix failing WITH_CRYPT_SERVICE test

aloisklink:
Adds CI tests for `linux-with-crypt`.

After adding these tests, I found that the tests were failing, since the `prepare_find_table` function was failing to find the `store` database, due to an `SQLITE_MISUSE` error.

I did a bit of refactoring to simplify the function and add error handling and the function and tests seem to work now.

**Draft until #227 is merged**


mereacre (approved):
Super. More tests.

PR #232, opened 2022-08-08
build: download uthash and utarray from upstream

aloisklink:
Instead of maintaining our own fork of uthash and utarray, we can download the upstream version.

This means we have less code to test, increasing code coverage (or at least it would if they weren't C/header-only macro functions).
It also makes it easier to update to a newer version of uthash the future (for example, we are now using v2.1.0, but v2.3.0 is now out)

I had a quick look at the git history for [utarray.h](https://github.com/nqminds/edgesec/commits/main/src/utils/utarray.h) and [uthash.h](https://github.com/nqminds/edgesec/commits/main/src/utils/uthash.h), and the only major thing that I could find is that [utarray.h was modified to use src/util/allocs.h](https://github.com/nqminds/edgesec/commit/e5820f46c0752613a3b6d5e82a23e66290b3d44b#diff-1e2475767cfc1ee3fc8eef260fa013dd1ebde4be2db9ed7cbd6b9c69259f0a9d).


mereacre (approved):
I like the idea. However, there's one main theoretical disadvantage that is if the author of the utarray decide to delete the repo.

aloisklink:
> I like the idea. However, there's one main theoretical disadvantage that is if the author of the utarray decide to delete the repo.

It's a bit too famous to delete :)

The uthash library is even built-in to Ubuntu, see https://packages.ubuntu.com/jammy/uthash-dev (`utarray.h` is one of the files it contains).

Apparently there's even a `/usr/lib/x86_64-linux-gnu/libut.a` file that makes compiling slightly faster.

PR #234, opened 2022-08-12
Test OpenWRT in CI

aloisklink:
Runs the OpenWRT tests in CI.

I had to slightly play around with some stuff to get OpenWRT tests working again.

Firstly, I had to add `LD_LIBRARY_PATH="${sourceDir}/build/${presetName}/lib/ubox/lib"`, see https://github.com/nqminds/edgesec/commit/ee360ebd6e586790c86fdc0245b54690519eb522 to get OpenWRT tests working properly, since even though we set RPATH on our tests, since the tests link to libuci, libuci is the one that needs RPATH to libubox.

This meant I had to modify the CI to use the presets listed in CMakePresets.json (see [ci: use buildDir from CMake presets](https://github.com/nqminds/edgesec/commit/df4bd607c8e66af522c62ef97321e37c3d4590ed)), then I had to add a preset for `openwrt` that set `LD_LIBRARY_PATH` (see [ci: add test config for openwrt](https://github.com/nqminds/edgesec/commit/ee360ebd6e586790c86fdc0245b54690519eb522))

Then, in order to fix the tests, I needed to:

- [test: add __wrap_uwrt_init_context test lib](https://github.com/nqminds/edgesec/commit/e967f82860ab1a240e3b16da5bc50cb2828fe023)
  The `__wrap_uwrt_init_context` test library can be used to wrap `uwrt_init_context()` so that it loads from the `./test/data/uci` folder.
- [tests(hostapd): fix WITH_HOSTAPD_UCI tests](https://github.com/nqminds/edgesec/commit/94f7930b18c1790c9d8ded726964b2d3a4a75b4f)
  The `run_ap_process()` function doesn't call `kill_process()` when WITH_HOSTAPD_UCI is defined.
  Additionally, no `hostapd_conf_file` is written, instead UCI handles writing the configuration file.
- [test: fix dnsmasq tests when using UCI mode](https://github.com/nqminds/edgesec/commit/451f5e1d92b7fc8f2ae59ad9ac90f32c0a54ffc2)
  When running dnsmasq in UCI mode:
  - The `dnsmasq_conf_path` property is ignored, as UCI decides where to store the configuration
  - `kill_process()` is not run when running `run_dhcp_process()`
  - `signal_dhcp_process()` does nothing

  Additionally, we need to make sure that `dhcp_conf.bridge_prefix` is correctly set, otherwise you'll get a bunch of undefined memory, and your tests will work 95% of the time, and fail 5% of the time. (I think it took me about an hour to debug this)
- [refactor: simplify uwrt_create_interface code](https://github.com/nqminds/edgesec/commit/0d16d12fc18473c7be697fb3bf9bc9819b48e89d)
    Some simple refactoring, just to avoid constantly doing `uwrt_set_property()`.
- [fix: set ifname in uwrt_create_interface()](https://github.com/nqminds/edgesec/commit/2ee2ff6f9f0fd5cd0ef4095c79ee899ab113ee78)
    The ifname field wasn't being set in `uwrt_create_interface`, which was causing the `uci_wrt.c` tests to fail.

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -32,16 +32,15 @@ jobs:
           sudo mk-build-deps --install --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes'  debian/control
       - name: Configure
         run: |
-          cmake -B build/ --preset "${{ matrix.cmake-preset }}" -DCONFIGURE_COVERAGE=BOOL:ON
+          cmake --preset "${{ matrix.cmake-preset }}" -DCONFIGURE_COVERAGE=BOOL:ON
       - name: Build
         run: |
-          # can't use --preset "${{ matrix.cmake-preset }}" since we want to specify a custom build dir
-          cmake --build build/ --parallel "$(($(nproc) + 1))"
+          cmake --build --preset "${{ matrix.cmake-preset }}" --parallel "$(($(nproc) + 1))"
       - id: test
         name: Test
         run: |
           if ctest --list-presets | grep -s "${{ matrix.cmake-preset }}"; then
-            ctest --test-dir build/ --preset "${{ matrix.cmake-preset }}" --output-on-failure
+            ctest --preset "${{ matrix.cmake-preset }}" --output-on-failure
             echo "::set-output name=tested::true"
           else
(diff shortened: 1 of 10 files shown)


mereacre (approved):
Great work.

PR #235, opened 2022-08-12
Tidies up the `uci_wrt.c` file

aloisklink:
~**Draft until #234 is merged.**~

Tidies up the `uci_wrt.c` file.

I've added two new functions:
  - `uwrt_set_properties()`
  - `uwrt_add_list_properties()`

They look like the following:

```c
/**
 * @brief Set multiple OpenWRT UCI properties at once
 * @param[in] ctx UCI context. The context ptr will be modified.
 * @param[in] properties Array of properties to set. Warning, strings may be
 * modified by UCI.
 * @retval  0 Success
 * @retval -1 Error
 */
int uwrt_set_properties(struct uci_context *ctx, UT_array *properties) {
  for (char *const *prop = (char **)utarray_front(properties); prop != NULL;
       prop = (char **)utarray_next(properties, prop)) {

    if (uwrt_set_property(ctx, *prop) < 0) {
      log_error("uwrt_set_property fail for %s", *prop);
      return -1;
    }
  }
  return 0;
}
```

Essentially, they just call `uwrt_set_property`/`uwrt_add_list_property` multiple times.

This means:
  - Less code reuse!
  - Code Coverage is happy since there are less branches.
  - Smaller code.
  - Bad news is that it's slightly slower.
    (I don't think the compiler is smart enough to optimise utarray loops)

I've also quickly added a simple test for `uci_wrt`, just to confirm that my changes don't cause any errors.


mereacre (approved):
Is it still in the draft mode?

aloisklink:
> Is it still in the draft mode?
Now that #234, I'll open it up as ready!

PR #238, opened 2022-08-23
ci: setup C CodeQL code quality scanner

aloisklink:
Sets up CodeQL to automatically scan for vulnerabilities in edgesec.

CodeQL is free for open-source repos on GitHub, so we might as well enable it!

If there are a bunch of false positives, we can always disable them later.

**This PR depends on #240, so you probably want to review that one first.**


mereacre (approved):
Great

PR #239, opened 2022-08-23
refactor: fix const char * warnings

aloisklink:
Converts some `char *` arrays into `const char *` arrays to fix compiler invalid const qualifier discardation errors.

E.g. converts: `char * example[] = {...};` to `const char *`

This also may fix a thread-safety bug in `ipaddr_modify()`.
`get_prefix()` modifies the input string temporarily.
However, even though the function changes the string back the original value, since edgesec uses multithreading, this may have caused an error.


tests/utils/test_nl.c
```diff
@@ -0,0 +1,66 @@
...
+  }
+  ptr->context = nl_init_context();
+  if (ptr->context == NULL) {
+    return -1;
+  }
+  *state = ptr;
+  return 0;
+}
+
+static int teardown(void **state) {
```
mereacre, line 31: Did you test if the interface is actually removed?

mereacre (approved):

tests/utils/test_nl.c
aloisklink: No need :)

I've only added a basic test that checks whether `nl_set_interface_ip()` fails if the interface doesn't exist, so there's actually no interface.

I couldn't figure out how to add a test that checks whether it actually works (I'm guessing we'd need to do a bunch of mocks).

tests/utils/test_nl.c
```diff
@@ -0,0 +1,66 @@
...
+  struct nl_test_state_t *ptr = *state;
+
+  expect_function_call(log_error);
+  expect_string(
+      __wrap_log_levels, error_message,
+      "ipaddr_modify error: could not find interface 'test_nl_interface_name'");
+  expect_function_call(log_error);
+  expect_string(__wrap_log_levels, error_message,
+                "nl_set_interface_ip error: ipaddr_modify failed with -1");
+  assert_int_equal(nl_set_interface_ip(ptr->context, test_nl_interface_name,
```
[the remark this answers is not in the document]
aloisklink: The test currently only checks for failure if the interface doesn't exist, so the right IP is never assigned.

I'm not 100% sure how to actually test if this function works (since tests need to run on CI too, and ideally without `sudo` access), but I guess it will take a lot of mocks.

PR #240, opened 2022-08-23
build(debian): link to sqlite3/libpcap shared libs

aloisklink:
Currently, libsqlite3 and libpcap are statically linked to edgesec, even when building a `.deb`.

Dynamic linking speeds up compilation, makes our package smaller (both on disk and in memory), and means sqlite3/libpcap can be updated to fix bugs without our package being updated.

The dynamically linked version of libsqlite3 is exactly the same as the statically linked version.

**However, the current statically linked version of libpcap is 1.10.1, which is newer than the dynamically linked libpcap 1.9.1** It seems to compile and run tests fine, though, so it probably means we aren't using any new libpcap features.


mereacre (approved):
Great work.

PR #242, opened 2022-08-23
Add Clang support to edgesec

aloisklink:
Adds initial basic support for compiling EDGESec with Clang.

Disables GNU/Clang C extensions by default on the edgesec code, except:
  - lib/libnetlink (taken from Linux kernel)
  - lib/libpcap (uses a bunch of BSD-style u_int)
  - any file that defined `#_GNU_SOURCE` (or similar)

Additionally, I've removed some GNU only features, like `C++11`-styled rawstrings.

Finally, I've added a CI job for Clang so it'll be tested. It even has code-coverage added!


aloisklink:
Huh, I've got no idea why adding Clang is changing the code-coverage values by so much (maybe it calculates lines slightly differently?). But Clang seems to be working, so I'll mark this PR as ready to review!

mereacre (approved):
Great

PR #243, opened 2022-08-26
build: convert LINK_FLAGS to target_link_options

aloisklink:
Instead of manually setting LINK_FLAGS for each target, we can use CMake's [target_link_options][1].

This has a few benefits:
- args can be scoped to be `INTERFACE|PUBLIC|PRIVATE`
- args are automatically deduplicated
- linker flags are portable for all compilers
  (e.g. these options will work both for Clang and GCC)

Downsides:
- Minimum CMake support is v3.13, which is fine, since our project minimum is currently v3.14

[1]: https://cmake.org/cmake/help/latest/command/target_link_options.html

### DIFF
--- a/tests/supervisor/CMakeLists.txt
+++ b/tests/supervisor/CMakeLists.txt
@@ -10,31 +10,32 @@ target_link_libraries(test_sockctl_server PRIVATE sockctl os log cmocka::cmocka)
 
 add_executable(test_cmd_processor test_cmd_processor.c)
 target_link_libraries(test_cmd_processor PRIVATE cmd_processor iptables log cmocka::cmocka)
-
-set(TEST_CMD_PROPS "-Wl,--wrap=write_socket_data -Wl,--wrap=accept_mac_cmd -Wl,--wrap=deny_mac_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=write_socket_data -Wl,--wrap=accept_mac_cmd -Wl,--wrap=deny_mac_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=add_nat_cmd -Wl,--wrap=remove_nat_cmd -Wl,--wrap=assign_psk_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=set_ip_cmd -Wl,--wrap=add_bridge_mac_cmd -Wl,--wrap=add_bridge_ip_cmd")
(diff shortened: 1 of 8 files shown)


tests/supervisor/CMakeLists.txt
```diff
@@ -10,31 +10,32 @@ target_link_libraries(test_sockctl_server PRIVATE sockctl os log cmocka::cmocka)
...
-set(TEST_CMD_PROPS "-Wl,--wrap=write_socket_data -Wl,--wrap=accept_mac_cmd -Wl,--wrap=deny_mac_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=write_socket_data -Wl,--wrap=accept_mac_cmd -Wl,--wrap=deny_mac_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=add_nat_cmd -Wl,--wrap=remove_nat_cmd -Wl,--wrap=assign_psk_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=set_ip_cmd -Wl,--wrap=add_bridge_mac_cmd -Wl,--wrap=add_bridge_ip_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=set_fingerprint_cmd -Wl,--wrap=query_fingerprint_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=clear_psk_cmd -Wl,--wrap=get_mac_mapper -Wl,--wrap=remove_bridge_cmd")
-set(TEST_CMD_PROPS "${TEST_CMD_PROPS} -Wl,--wrap=clear_bridges_cmd -Wl,--wrap=subscribe_events_cmd -Wl,--wrap=register_ticket_cmd")
+target_link_options(test_cmd_processor PRIVATE
+  "LINKER:--wrap=write_socket_data,--wrap=accept_mac_cmd,--wrap=deny_mac_cmd"
+  "LINKER:--wrap=write_socket_data,--wrap=accept_mac_cmd,--wrap=deny_mac_cmd"
```
aloisklink, line 15: It looks like there's a duplicate line here.

CMake will automatically duplicate these flags, which might be bad if these are purposely duplicated (I think it's just accidental and we can remove one of these lines)

```suggestion
```

mereacre (approved):
Great as it simplifies the cmake files.

aloisklink:
> Great as it simplifies the cmake files.

True :) Although the Clang support is pretty nice too!

PR #244, opened 2022-08-26
test: test for NULL string in string_append_char

aloisklink:
Adds a test case for `string_append_char` when the input string is null.


AshleySetter (approved):
:+1:

PR #245, opened 2022-08-30
build: fix unknown autoconf os error message

aloisklink:
Fix the error message that is thrown when the autoconf OS cannot be converted from the CMAKE_HOST_SYSTEM_NAME var.

In CMake, variables always have to be quoted in `${VAR}`, using just `$VAR` is not enough.


AshleySetter (approved):
Good catch! :eyes:

PR #246, opened 2022-08-30
refactor: replace <linux/if.h> with <net/if.h>

aloisklink:
The `#include <linux/if.h>` header is Linux exclusive.

However, the `#include <net/if.h>` header is part of the POSIX standard, so it's more cross-platform compatible.

Additionally, I've tried to delete this include from files that don't need it.


mereacre (approved):
Great

Issue #252, opened 2022-09-05
Intermittent failing CI due to `Failed to execute statement: disk I/O error`


aloisklink:
**Describe the bug**

GitHub Actions CI tests are failing intermittently due to:

```
 5/35 Test  #4: test_sqlite_header ...............***Failed    0.01 sec
[==========] Running 1 test(s).
[ RUN      ] test_init_sqlite_header_db
2022-09-05 08:31:12.358  ERROR sqliteu.c:23: Failed to execute statement: disk I/O error
2022-09-05 08:31:12.358  ERROR sqlite_header.c:614: execute_sqlite_query fail: CREATE TABLE IF NOT EXISTS eth (timestamp INTEGER NOT NULL, id TEXT NOT NULL, caplen INTEGER, length INTEGER, ifname TEXT, ether_dhost TEXT, ether_shost TEXT, ether_type INTEGER, PRIMARY KEY (timestamp, id));
```

Interestingly, our CI runs tests twice. Once for testing, and once for code coverage. This bug seems to only have on the 2nd run, so I'm guessing there may be some kind of race condition (or maybe GitHub Actions has an unusual file-system)

My gut feeling is that a `sleep 5s` between tests runs should prevent it.

**To Reproduce**

Intermittent, so we can't easily reproduce.

**Expected behavior**

CI tests should be consistent.

aloisklink:
Reopening, this still seems to be happening, see https://github.com/nqminds/edgesec/actions/runs/3089564730/jobs/4997321185

aloisklink:
Hi @mereacre, I'm guessing this issue should be fixed by https://github.com/nqminds/edgesec/commit/cecb5e4420f2f6a4d7fea96fd0d474ce5510cdb0.

If that commit doesn't fix this issue, feel free to re-open this.

PR #254, opened 2022-09-05
ci: fix intermittent disk I/O error

aloisklink:
GitHub Actions CI tests are failing intermittently due to an SQLite `Failed to execute statement: disk I/O error`.

I'm pretty sure this is due to some sort of race condition, either with:
  - a) our thread code
  - b) GitHub Actions file-system

This commit adds a `sleep 5` between the CI tests and the coverage tests in an attempt to prevent this error.

Fixes https://github.com/nqminds/edgesec/issues/252 (hopefully at least!)

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -63,6 +63,8 @@ jobs:
         # code coverage only works if ctest works
         if: steps.test.outputs.tested == 'true'
         run: |
+          # Should prevent `Failed to execute statement: disk I/O error` errors
+          sleep 5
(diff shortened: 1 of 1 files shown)


mereacre:
Yes, it might be the capture threads as all of them are trying to write to the same db.

mereacre (approved):

aloisklink:
I tried testing locally with the following, and I couldn't reproduce it locally. Maybe it might be because GitHub Actions is pretty slow.

```bash
for i in {1..50}; do
  echo "$i";
  ctest --preset linux-with-example-middlewares --output-on-failure;
done
```

PR #255, opened 2022-09-05
refactor: remove unnecessary middlewares includes

aloisklink:
Including `<netinet/ip.h>` is causing issues on FreeBSD.

These includes aren't actually needed, so we might as well get rid of them and speed up compilation slightly.


mereacre (approved):
Great.

PR #116, opened 2022-04-12
Sort `#include`'s in `clang-format`

aloisklink:
This PR activates the `#include` sorting part of `clang-format`, since in C and C++, **compiling behaviour might change when `#include` order is changed** (e.g. `cmocka.h` must be `#include`-d after `<stdint.h>`

PR #257, opened 2022-09-05
refactor: add #include <stdint.h> before cmocka

aloisklink:
CMocka requires `#include <stdint.h>` to be included before every `#include <cmocka.h>`. On Linux, some of the other headers include `#include <stdint.h>` for us, but on FreeBSD, we need to manually include it.

### DIFF
--- a/tests/ap/test_ap_service.c
+++ b/tests/ap/test_ap_service.c
@@ -8,11 +8,11 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <stddef.h>
-#include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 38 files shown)


tests/ap/test_ap_service.c
```diff
@@ -8,11 +8,11 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <stddef.h>
-#include <stdint.h>
```
aloisklink, line 11: I've moved `#include <stdint.h>` to be right next to `#include <cmocka.h>` to make Regex searching easier. This should also be more obvious once #116 gets merged.

PR #259, opened 2022-09-05
refactor(wpabuf): simplify byte swaps

aloisklink:
Cleans up the `wpabuf.h` file to improve support for non Linux OSes like FreeBSD.

As of glibc 2.19, the following macros are defined in `endian.h`:
  ```
  htobe16(), htole16(), be16toh(), le16toh(), htobe32(), htole32(),
  be32toh(), le32toh(), htobe64(), htole64(), be64toh(), le64toh()
  ```

These macros are also defined in FreeBSD, however they are located in the `<sys/endian.h>` file.

### DIFF
--- a/src/radius/wpabuf.h
+++ b/src/radius/wpabuf.h
@@ -1,15 +1,12 @@
-/*
- * Dynamic data buffer
- * Copyright (c) 2007-2012, Jouni Malinen <j@w1.fi>
- *
- * This software may be distributed under the terms of the BSD license.
- * See README for more details.
- */
-
 /**
+ * @brief Dynamic data buffer.
  * @file wpabuf.h
  * @author Jouni Malinen
- * @brief Dynamic data buffer.
+ * @author Alois Klink
+ * @copyright
+ * SPDX-FileCopyrightText: Copyright (c) 2007-2012, Jouni Malinen <j@w1.fi>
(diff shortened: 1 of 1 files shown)


src/radius/wpabuf.h
```diff
@@ -22,66 +19,20 @@
...
-#define le_to_host32(n) bswap_32(n)
-#define host_to_le32(n) bswap_32(n)
-#define be_to_host32(n) (n)
-#define host_to_be32(n) (n)
-#define le_to_host64(n) bswap_64(n)
-#define host_to_le64(n) bswap_64(n)
-#define be_to_host64(n) (n)
-#define host_to_be64(n) (n)
-#ifndef WORDS_BIGENDIAN
-#define WORDS_BIGENDIAN
```
aloisklink, line 80: This `#define` is the same as the one in https://github.com/nqminds/edgesec/blob/2ae010960fb1ad48a38a95ec2e7331285e63f82d/src/radius/md5_internal.c#L66,
however `wpabuf.h` is not `#include`-ed by `md5_internal.c`, so it's safe to remove.

mereacre: As long as the radius test rusn. Should be fine.

mereacre (approved):

PR #260, opened 2022-09-05
test: remove incorrect error_t type

aloisklink:
These functions return `int`, and `error_t` is only a Linux type, and not supported on FreeBSD.


mereacre (approved):

mereacre:
Very strange. I'm not even aware of `error_t` type.

PR #261, opened 2022-09-06
refactor: use POSIX standard udphdr/tcphdr

aloisklink:
Refactors the `packet_decoder` so that the POSIX standard values for the `udphdr` and `tcphdr` struct are used.

On Linux, these structs are defined as a union, so that the POSIX standard can also be used. However, other platforms like FreeBSD only have the POSIX standard.


mereacre (approved):
Great

Issue #170, opened 2022-06-20
Eloop termination


mereacre:
New future to terminate eloop:
 - Add eventfd https://man7.org/linux/man-pages/man2/eventfd.2.html
 - Add mutex

aloisklink:
We can't use `eventfd` since it's Linux only, and won't be supported on FreeBSD.

If we do add mutex/locks, I'd recommend using standard C `threads.h` locks, instead of using `pthreads` (`pthreads` is supported on both FreeBSD and Linux, but both might have some non-POSIX extensions)

PR #265, opened 2022-09-09
Use eloop from hostapd

aloisklink:
The `src/utils/eloop.c` file is only Linux compatible, and is not compatible with FreeBSD.

However, the eloops.c/.h file from the 2.10 HostAPD source code contains FreeBSD support. Instead of copying the file into our git repo, we can instead download it, and manually apply patches to the file.

This should make it much easier to upgrade to v2.11 of HostAPD, or to take in any other upstream improvements to eloop.

Unfortunately, a the `git history` of `src/utils/eloop.c` is a bit missing (since a lot of modifications were already in https://github.com/nqminds/edgesec/commit/2a18535cb57f7059dba72eb50490bda587291fe8), but I tried to do the bare minimum amount of changes to `eloop` to get it to compile/tests to run.

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -66,7 +66,7 @@ jobs:
         name: Test
         run: |
           if ctest --list-presets | grep -s "${{ matrix.cmake-preset }}"; then
-            ctest --preset "${{ matrix.cmake-preset }}" --output-on-failure
+            # ctest --preset "${{ matrix.cmake-preset }}" --output-on-failure
             echo "::set-output name=tested::true"
           else
             echo "::set-output name=tested::false"
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ dkms.conf
 # get re-commited
 /lib/libmicrohttpd-0.9.72
 
+compile_commands.json
+Testing/**
+
 apps/**/node_modules
 docs/**/node_modules
 docs/docusaurus/.eslintcache
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,9 +96,6 @@ option(USE_GENERIC_IP_SERVICE "Use generic ip service" OFF)
 
 option(USE_CRYPTO_SERVICE "Use the crypto service" OFF)
 cmake_dependent_option(BUILD_OPENSSL_LIB "Build OpenSSL" ON USE_CRYPTO_SERVICE OFF)
-if (USE_CRYPTO_SERVICE)
-  add_compile_definitions(TEST_CRYPTO_SERVICE_KEY="12345")
-endif ()
 
 option(USE_ZYMKEY4_HSM "Use the Zymkey4 HSM" OFF)
 option(CONFIGURE_COVERAGE "Configure for code coverage (requires lcov)" OFF)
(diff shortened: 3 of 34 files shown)


lib/eloop/patches/0003-Bump-timeout-parameters-from-int-to-long.patch
```diff
@@ -0,0 +1,87 @@
+From 03270444add875925fef1e2cb97647f45a394d30 Mon Sep 17 00:00:00 2001
+From: Alois Klink <alois@nquiringminds.com>
+Date: Tue, 6 Sep 2022 18:34:47 +0100
+Subject: [PATCH 3/3] Bump timeout parameters from int to long
+
+I'm not 100% sure why this was done.
+I'm guessing that on some platforms, the max int value is only
+32768, and that may cause overflow.
```
aloisklink, line 8: Is this description correct?

I tried to understand what https://github.com/nqminds/edgesec/commit/00d465d8705fa439b52a513541dc03c1f4231e4d was doing, but I'm not 100% sure if I was right.

mereacre:
Before we start working on this pull request we need to add loads of tests for eloop.

The reason is that I completely modified the eloop, i.e., it doesn't use any global variables and instead it uses a context which the user can allocate on the heap.

aloisklink:
> Before we start working on this pull request we need to add loads of tests for eloop.

> The reason is that I completely modified the eloop, i.e., it doesn't use any global variables and instead it uses a context which the user can allocate on the heap.

:+1:

I added your changes in https://github.com/nqminds/edgesec/blob/78cb3006dafb782f368868b3aafd2f8e62243be5/lib/eloop/patches/0002-Allow-multiple-eloops-to-run-at-same-time.patch

The `eloop.h` file is the same as your version in `src/utils/eloop.h`, but the `eloop.c` implementation might be quite different (although I did try to copy as much as possible). I basically just added `eloop` as a parameter to every function.

mereacre:
I already added tests in eloop-test branch. Can you update the branch with your tests?

aloisklink:
I've removed the test commits from this PR, since they were added in PR #275 instead.

Tests pass on FreeBSD using kqueue :rocket:

mereacre:
@aloisklink before we merger the eloop we need to add several integration test, i.e., tests that do not depend on presence of the hardware.

aloisklink:
> - We need to simulate the hostapd

This one might be hard to test, especially since we need to be able to run this test somewhere we:

a) don't have root access, and
b) don't have `sudo` access.

Maybe this isn't necessary: currently we have tests for every function in `src/utils/eloop.h` except:

- eloop_register_sock()
- eloop_unregister_sock()
- eloop_wait_for_read_sock()
- eloop_sock_requeue()

(some of these might be called internally by `eloop_register_read_sock()`, which we do test)

> - We need to run a radius client in test mode

Isn't this what `test_radius_server` does? If no, can you make a GitHub Issue with what you want tested? Since you know more than I what these files actually do.

> - We need to implement eloop termination see https://github.com/nqminds/edgesec/issues/170

:+1: It's probably easier to get #170 working before merging this PR.

Just by the way, remember that without merging this PR, we can't get EDGESec compiled on FreeBSD.

aloisklink:
I've `git rebase`-d onto the latest `main` branch, just to hopefully simply things after #339 is merged too.

mereacre (approved):
Great work. Hopefully the tests pass.

PR #266, opened 2022-09-09
test: fix invalid test variable type

aloisklink:
My fault, I accidentally wrote `const` instead of `char`, and C implicitly assumed that it was an `int` pointer.

Maybe it's worth setting `-Werror` in our code so that all C warnings become errors.


mereacre (approved):

mereacre:
I added once (a year or so ago) the -Werror. There was a bunch of errors. But we should add it again. I think this time there will be less errors.

PR #276, opened 2022-09-16
Cleanup sockctl tests

aloisklink:
It's probably easier to review each commit one by one!

The main point of this PR is to remove the use of `tempnam` from this project, as it is deprecated and the compiler/linker throws a warning on use.

The first line of the description in the [tempnam manpage](https://manpages.ubuntu.com/manpages/jammy/en/man3/tempnam.3.html) says:

> _Never use this function_.  Use mkstemp(3) or tmpfile(3) instead.

I've also used `teardown()` functions to automatically cleanup all of the sockctl temporary files, so that my `/tmp` folder isn't filled with junk! 

In commit [test(sockctl): refactor and cleanup tests](https://github.com/nqminds/edgesec/commit/77e6fd332cfe1d55698e6c1788eb717e655c5f11), I've also done some general cleanup.

### DIFF
--- a/tests/supervisor/test_sockctl_server.c
+++ b/tests/supervisor/test_sockctl_server.c
@@ -130,11 +130,7 @@ static void test_create_domain_client(void **state) {
 static void test_read_domain_data_s(void **state) {
   struct test_state *test_state = *state;
 
-  struct client_address addr;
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Very good.

PR #278, opened 2022-09-19
Add fallback for unsupported abstract sockets

aloisklink:
Currently, [`create_domain_client(NULL)`](https://nqminds.github.io/edgesec/sockctl_8c.html#ab74693481cda01c302ed453c910a4096) creates a new temporary _abstract_ Unix domain socket:

> **Abstract sockets**
>
> [...]
> Abstract sockets automatically disappear when  all  open  references  to  the  socket  are closed.
>
> The abstract socket namespace is a nonportable Linux extension.

(from https://manpages.ubuntu.com/manpages/jammy/en/man7/unix.7.html)

Since abstract sockets are "a notportable Linux extension", they are not supported by FreeBSD.

Because of this, I've instead replaced this code with creating a new _pathname_ Unix domain socket, where the socket is located in `/tmp/edgesec.XXXXXX/client-socket.sock`, where `/tmp/edgesec.XXXXXX` is made with `mkdtemp`.

**By default, this code is only used on non-Linux platforms, like FreeBSD**.

I then added a new function called `close_domain_socket()` that:
- closes Unix domain sockets, and
- `unlink()` the `pathname` (only for `_pathname_` Unix domain sockets), and
- `rmdir()` the `/tmp/edgesec.XXXXXX` (only if the socket was created by `create_domain_client(NULL)` (I'm using a `uthash` map to confirm this)).

### DIFF
--- a/src/utils/sockctl.c
+++ b/src/utils/sockctl.c
@@ -27,8 +27,7 @@
 #define DOMAIN_REPLY_TIMEOUT 10
 
 void init_domain_addr(struct sockaddr_un *unaddr, const char *addr) {
-  os_memset(unaddr, 0, sizeof(struct sockaddr_un));
-  unaddr->sun_family = AF_UNIX;
+  *unaddr = (struct sockaddr_un){.sun_family = AF_UNIX};
   os_strlcpy(unaddr->sun_path, addr, sizeof(unaddr->sun_path));
 }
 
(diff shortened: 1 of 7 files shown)


src/utils/sockctl.c
```diff
@@ -96,6 +96,33 @@ int create_domain_server(const char *server_path) {
...
+    log_errno("Failed to getsockname for unix domain socket %d",
+              unix_domain_socket_fd);
+    return -1;
+  }
+  if (sockaddr.sun_family != AF_UNIX) {
+    log_error("Socket %d is not a unix domain socket, instead it's a %d",
+              unix_domain_socket_fd, sockaddr.sun_family);
+    return -1;
+  }
+  if (address_len >=
```
mereacre, line 192: Can you define a function for instance `is_unix_domain_socket` that returns 1 if it is and 0 if it is not and -1 on error?

src/utils/sockctl.c
```diff
@@ -23,13 +23,83 @@
...
-
+#define TMP_UNIX_SOCK_FOLDER_TEMPLATE "/tmp/edgesec.XXXXXX"
+#define TMP_UNIX_SOCK_NAME "client-socket" SOCK_EXTENSION
 #define DOMAIN_REPLY_TIMEOUT 10
 
 void init_domain_addr(struct sockaddr_un *unaddr, const char *addr) {
   *unaddr = (struct sockaddr_un){.sun_family = AF_UNIX};
   os_strlcpy(unaddr->sun_path, addr, sizeof(unaddr->sun_path));
 }
+struct tmp_domain_socket {
```
mereacre, line 34: Could you please explain this structure?

aloisklink: How are these doxygen docs? (added in https://github.com/nqminds/edgesec/pull/278/commits/b90bc2dd16e7b9ba430420a2dd79b1c7ae4355f7)

Essentially, it's a hashmap struct that stores all the temporary unix domain sockets (and the folder they are in) created by `create_tmp_domain_socket_path()` so that the sockets and the folders can be properly deleted by `cleanup_tmp_domain_socket_path()`

https://github.com/nqminds/edgesec/blob/b90bc2dd16e7b9ba430420a2dd79b1c7ae4355f7/src/utils/sockctl.c#L38-L54

src/utils/sockctl.c
mereacre: You are right. It is not worth it.

aloisklink: Yeah, that's the annoying bit about C. Because you need to explicitly handle every return code, and almost every single function can return an error, little helper functions are sometimes not worth it (although I've heard Rust has similar error handling semantics).

src/utils/sockctl.c
mereacre: We are using here a global variable. In general I prefer not to use any in the edgesec code. I think we should refactor and add things like context or similar. There's also the problem of thread safety. We haven't done any locking for the hash string. Check the thread safety section in https://troydhanson.github.io/uthash/userguide.html

aloisklink: > We are using here a global variable. In general I prefer not to use any in the edgesec code.

Ahhh, that will make things difficult. I agree, global variables are a nightmare. I only implemented it that way to change as little as I could as possible.

It's a pity that in C you often have to use global variables, especially for stuff like signal handling.

Hmmmmm, in that case, I need to think of the best way of how to handle cleanup. How about:
  - Return a `struct unix_socket {int fd; (void *)(void) destructor;}` from `create_domain_client()`, so we can return a custom `destructor()` function (C++ style)
  - Just ignore cleaning up the domain sockets.
    - They're only used in FreeBSD
    - If we stick them all into a `/tmp/edgesec/tmp-sockets.XXXXXX/*`, we can just not bother cleaning them up, and as long it takes at years for 1 million tmp files to be created, FreeBSD will eventually shutdown and delete the `/tmp` folder.

mereacre: I think the best answer is to ignore cleaning up the tmp files.

mereacre: Could you please add a `log_warn` or similar saying which files are being deleted?

mereacre (approved):

src/utils/sockctl.c
aloisklink: Added in https://github.com/nqminds/edgesec/pull/278/commits/ae16f1c00642f322e1b870a48f8100ea833c7422

I've made it is only a `log_debug()`. This is because if we do try to delete the wrong folder, chances are that it's not empty and so `rmdir()` will have a `ENOTEMPTY` (dir not empty) error, and we'll get a `log_errno()` message.

PR #279, opened 2022-09-19
Standardise `test_run_commands` test

aloisklink:
test_run_command() expects that the output of `uname -s` is always "Linux\n". However, this isn't right for two reasons:
- That string isn't guaranteed to be NUL-terminated, so one day it might instead be `Linux\n<some_junk_data_here>`.
- On other platforms like FreeBSD, `uname -s` does not print Linux.

Because of this, I've changed the test command to print "Hello World!", and to handle the NUL-terminator and some other stuff.


mereacre (approved):
Great

PR #280, opened 2022-09-19
refactor(middlewares): fix const stype difference

aloisklink:
`ltype` was made const in commit 669377818af88075f604e71224a703820e30704b but some middlewares missed the change.

Fixes: 669377818af88075f604e71224a703820e30704b


mereacre (approved):
Great

PR #281, opened 2022-09-19
feat(os): add copy_argv() function

aloisklink:
C programs such as `run_command()` accept argv as an `char *`, not a `const char *`. This is because the `main()` function of programs is explicitly allowed to modify the data in `argv()`.

This can cause issues, since we declare arrays using string literals, (e.g. `char * argv[] = {"hello", "world", NULL};`), and modifying string literals is undefined behaviour.

To get around this, we can just make a mofiiable copy of argv before passing it to run_command().

There is a warning we can enable that will check for this (`-Wwrite-strings`), but it creates 100s of warnings throughout our code-base, so it's something we should slowly fix.

### DIFF
--- a/src/utils/os.c
+++ b/src/utils/os.c
@@ -288,6 +288,10 @@ int read_command_output(int fd, process_callback_fn fn, void *ctx) {
 
 char **copy_argv(const char *const argv[]) {
   // calculate the length of argv (without NULL terminator)
+  if (argv == NULL) {
+    log_error("argv param is NULL");
+    return NULL;
+  }
   size_t argc = 0;
   while (argv[argc] != NULL) {
     argc++;
--- a/src/utils/os.c
+++ b/src/utils/os.c
@@ -287,11 +287,12 @@ int read_command_output(int fd, process_callback_fn fn, void *ctx) {
 }
 
 char **copy_argv(const char *const argv[]) {
-  // calculate the length of argv (without NULL terminator)
   if (argv == NULL) {
     log_error("argv param is NULL");
     return NULL;
   }
+
+  // argc is the length of argv (excluding the NULL terminator)
   size_t argc = 0;
   while (argv[argc] != NULL) {
     argc++;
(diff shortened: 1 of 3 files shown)


src/utils/os.c
```diff
@@ -286,6 +286,40 @@ int read_command_output(int fd, process_callback_fn fn, void *ctx) {
   return count;
 }
 
+char **copy_argv(const char *const argv[]) {
+  // calculate the length of argv (without NULL terminator)
```
mereacre, line 290:
```c
if (argv == NULL) {
  log_error("argv param is NULL");
  return NULL;
}
```

src/utils/os.c
```diff
@@ -286,6 +286,40 @@ int read_command_output(int fd, process_callback_fn fn, void *ctx) {
...
+    strings_length += (strlen(argv[i]) + 1);
+  }
+
+  // first part of malloc-d data holds argv, second part stores string data
+  char **const argv_copy = (char **)malloc(argv_array_size + strings_length);
+  if (argv_copy == NULL) {
+    log_errno("Failed to malloc %d bytes", argv_array_size + strings_length);
+    return NULL;
+  }
+  char *const argv_string_buffer = &((char *)argv_copy)[argv_array_size];
```
mereacre, line 309: Not clear where `argv_copy[0]` is assigned.

mereacre (commented):
Need tests for `copy_argv`

src/utils/os.c
aloisklink: Thanks for the suggestion.

By the way, you can use the following syntax so that GitHub has the fancy "Commit suggestion":

````markdown
```suggestion
you're suggestion here
```
````

mereacre: Thanks for reminding me. I always forget about this trick.

src/utils/os.c
aloisklink: `argv_copy[0]` is assigned a few lines down: https://github.com/nqminds/edgesec/blob/e6bb700294f822654c55b1e06afc70ec01315441/src/utils/os.c#L314

The first half of the `malloc()`-ed data is filled `char*` (string pointers).
The second half of the `malloc()`-ed data is where the string buffer/contents are.

Essentially, if you copy `const char * argv[] = {"Hello", "World!", NULL};`, the data structure of `argv_copy` looks something like the following:
 
| **Address** |   0-7 |  8-15 | 16-23 |   24-29 |    30-36 |
| ----------- | ----: | ----: | ----: | ------- | -------- |
|  **CType**  | char* | char* | char* |  char[] |   char[] |
|  **Value**  |    24 |    30 |  NULL | "Hello" | "World!" |

I've added some more comments, since I had to make the data structure a bit confusing so that a single `free()` will delete all the data, see https://github.com/nqminds/edgesec/blob/430604f5b03b41a36d87ab265c5247fb711c41b4/src/utils/os.c#L332-L336

aloisklink:
Tests added in https://github.com/nqminds/edgesec/pull/281/commits/430604f5b03b41a36d87ab265c5247fb711c41b4!

mereacre (approved):
Thanks. That's clear.

PR #292, opened 2022-09-20
Error on all C/C++ compiler warnings in `src/` and `tests/`

aloisklink:
Enables the GCC/Clang `-Werror` flag on everything in the src/ and tests/ folder.

It's only enabled if `-DCMAKE_BUILD_TYPE=Debug`. This is so that any 3rd parties using edgesec (for example https://github.com/nqminds/manysecured-openwrt-packages or even the `.deb` build) should still be able to compile edgesec, even if they have a newer version of GCC/Clang that adds new warnings.

These warnings are usually pretty useful, so we probably want to fix them all. If we ever want to disable this, we can do:

```cmake
add_compile_options(-Wno-error=foo) # too many -Wfoo warnings
```

Or better yet, we can use [`target_compile_options`](https://cmake.org/cmake/help/latest/command/target_compile_options.html) to disable them just in one file:

```cmake
target_compile_options(my_library_with_warnings
  PUBLIC -Wno-error=foo # error is also disabled for all dependencies of this lib (e.g. error is in header file)
  PRIVATE -Wno-error=foo # error is not disable for dependencies of this lib (e.g. error is only in .c source file
)
```

There was only one minor warning left to fix in our code, see [test(os): fix const correctness argv warning](https://github.com/nqminds/edgesec/commit/c7e3239f93c589e639f47633083c15f6d784a28e).

(OpenSSL, `lib/libnetlink`, `sqlite3` and `libpcap` print a lot of warnings, but since they're not in `src/` and `tests/`, they won't be converted to errors.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,3 +1,6 @@
+# Error on all C/C++ warnings in the src/ directory if making a Debug build
+add_compile_options($<$<CONFIG:Debug>:-Werror>)
+
 add_subdirectory(utils)
 add_subdirectory(supervisor)
 add_subdirectory(ap)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,3 +1,6 @@
+# Error on all C/C++ warnings in the tests/ directory if making a Debug build
+add_compile_options($<$<CONFIG:Debug>:-Werror>)
+
 include_directories (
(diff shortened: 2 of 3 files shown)


mereacre (approved):
Great. We need to document(write as comment) in the main CMakeLists file how to disable teh error for specific targets.

PR #294, opened 2022-09-23
Add Tap Middleware for injecting packets into interface

mereacre:
- Created a new middleware called "tap"  that injects packets into an interface. The middleware does not create the interface where the packets are injected. It is the responbility of the system configuration script
 - Added a new parameter to `config.ini` section `capture` called `params`, where the dev can assign custom parameters for middleware. It will be the responsibility of the interface to process it.
 - Added example script to create a tap interface based on [GRE](https://en.wikipedia.org/wiki/Generic_Routing_Encapsulation) tunnel protocol. That is when the packets are injected the GRE sends the packets to a remote destination.  
 - Created a new middleware called `linux-with-tap`.


src/capture/middlewares/tap_middleware/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include_directories (
+  "${PROJECT_SOURCE_DIR}/src"
+)
+
+add_library(tap_middleware tap_middleware.c)
+target_include_directories(tap_middleware PRIVATE ${PROJECT_BINARY_DIR})
+target_link_libraries(tap_middleware PUBLIC PCAP::pcap PRIVATE pcap_service log os SQLite::SQLite3)
```
aloisklink, line 7:
```suggestion
target_link_libraries(tap_middleware PUBLIC SQLite::SQLite3 pcap_service eloop PRIVATE allocs os log squeue)
```

`tap_middleware.h` includes `middleware.h`, which includes `SQLite::SQLite3 pcap_service eloop`. Ideally, we'd probably want to make a `middleware` CMake target, but that can be a future refactor.

src/capture/pcap_service.c
```diff
@@ -112,16 +112,13 @@ int run_pcap(char *interface, bool immediate, bool promiscuous, int timeout,
   struct bpf_program fp;
   struct pcap_context *ctx = NULL;
 
-  if (!find_device(interface, &net, &mask)) {
-    log_error("find_interfaces fail");
-    return -1;
+  if (find_device(interface, &net, &mask) > 0) {
+    bit32_2_ip((uint32_t)net, ip_str);
+    bit32_2_ip((uint32_t)mask, mask_str);
```
aloisklink, line 117: Worth doing error checking here?

```suggestion
    if(bit32_2_ip((uint32_t)net, ip_str) == NULL) {
      log_errno("Converting %d to IP failed", net);
      return -1;
    }
    if(bit32_2_ip((uint32_t)mask, mask_str) == NULL) {
      log_errno("Converting %d to IP mask failed", mask);
      return -1;
    }
```

src/capture/pcap_service.c
```diff
@@ -112,16 +112,13 @@ int run_pcap(char *interface, bool immediate, bool promiscuous, int timeout,
   struct bpf_program fp;
   struct pcap_context *ctx = NULL;
 
-  if (!find_device(interface, &net, &mask)) {
-    log_error("find_interfaces fail");
-    return -1;
+  if (find_device(interface, &net, &mask) > 0) {
```
aloisklink, line 115: Do we need to check for error if `find_device(interface, &net, &mask) < 0`?
If no, maybe we should add a comment saying that we're ignoring the error.

src/capture/pcap_service.c
```diff
@@ -223,6 +220,17 @@ int dump_file_pcap(struct pcap_context *ctx, char *file_path,
   return 0;
 }
 
+int inject_pcap(struct pcap_context *ctx, uint8_t *packet, size_t size) {
+  int sent;
+
+  if ((sent = pcap_inject(ctx->pd, packet, size)) < 0) {
```
aloisklink, line 226: We should always initialise variables at declaration if we can. This follows the Google style guide, so we can call ourselves Google-C++-engineers https://google.github.io/styleguide/cppguide.html#Local_Variables :wink: 
```suggestion
  int sent = pcap_inject(ctx->pd, packet, size);

  if (sent < 0) {
```

src/config.c
```diff
@@ -554,6 +553,13 @@ bool load_capture_config(const char *filename, struct capture_conf *config) {
   os_strlcpy(config->filter, value, MAX_FILTER_SIZE);
   os_free(value);
 
+  // Load middleware params
+  value = os_zalloc(INI_BUFFERSIZE);
+  ini_gets("capture", "middlewareParams", "", value, INI_BUFFERSIZE, filename);
+
+  os_strlcpy(config->middleware_params, value, MAX_MIDDLEWARE_PARAMS_SIZE);
+  os_free(value);
```
aloisklink, line 561: Instead of allocating on the heap, (and having to check whether `os_zalloc()` fails), we could just allocated a buffer on the stack instead, which will make memory management so much easier.

```suggestion
  char ini_buffer[INI_BUFFERSIZE] = "";
  ini_gets("capture", "middlewareParams", "", ini_buffer, INI_BUFFERSIZE, filename);

  os_strlcpy(config->middleware_params, ini_buffer, MAX_MIDDLEWARE_PARAMS_SIZE);
```

aloisklink (commented):
Looks good to me!

I've got a bunch of small changes/comments, but none of them are blocking, so you're welcome to merge if you disagree with any of them.

As a side-note, it might be worth setting up https://github.com/shellcheck-py/shellcheck-py so that it automatically scans our shell scripts for basic errors/issues.

src/capture/middlewares/tap_middleware/tap_middleware.c
```diff
@@ -0,0 +1,118 @@
...
+  if (context != NULL) {
+    struct pcap_context *pctx = (struct pcap_context *)context->mdata;
+    if (pctx != NULL) {
+      close_pcap(pctx);
+    }
+    os_free(context);
+  }
+}
+
+struct middleware_context *init_tap_middleware(sqlite3 *db, char *db_path,
```
aloisklink, line 43: **required**: I believe this is what is causing your CI to fail.

```suggestion
struct middleware_context *init_tap_middleware(sqlite3 *db, const char *db_path,
```

deployment/scripts/gre_tunnel.sh
```diff
@@ -0,0 +1,22 @@
+#ubuntu@host-2:~$ sudo ip tunnel add tun0 mode gre remote 10.131.73.9 local 10.131.73.16 dev eth0
+#ubuntu@host-2:~$ sudo ip addr add 172.17.0.2/24 dev tun0
+#ubuntu@host-2:~$ sudo ip link set tun0 up
+#!/usr/bin/env bash
+
+TAP_IFACE=$1
+TAP_IP=$2
+LOCAL_IP=$3
+REMOTE_IP=$4
+INTERFACE=$5
```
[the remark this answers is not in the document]
mereacre: Thanks.

src/capture/middlewares/tap_middleware/tap_middleware.c
```diff
@@ -0,0 +1,118 @@
...
+    log_error("run_pcap fail");
+    free_tap_middleware(context);
+    return NULL;
+  }
+
+  context->mdata = (void *)pctx;
+  return context;
+}
+
+int process_tap_middleware(struct middleware_context *context, char *ltype,
```
aloisklink, line 90: Whoop, sorry, it's this line you need to make `const`, not the other line! My fault!

```suggestion
int process_tap_middleware(struct middleware_context *context, const char *ltype,
```

src/capture/pcap_service.c
mereacre: Added comment.

deployment/scripts/mirror_port.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# egress
+tc qdisc add dev $SRC_IFACE handle 1: root prio
+tc filter add dev $SRC_IFACE parent 1: \
+          protocol all \
+          u32 match u8 0 0 \
+          action mirred egress mirror dev $DST_IFACE
+
+echo "Hit Ctrl-C or kill this session to end port mirroring"
+sleep infinity
```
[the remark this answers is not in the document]
mereacre: This is only a test script. Will not be used in production.

src/capture/middleware.h
```diff
@@ -23,6 +23,7 @@ struct middleware_context {
   struct eloop_data *eloop;
   struct pcap_context *pc;
   void *mdata;
+  char *params;
```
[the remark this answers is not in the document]
mereacre: I added a comment. `params` points to `char params[MAX_MIDDLEWARE_PARAMS_SIZE];`.

aloisklink (commented):
Looks good to me!

Please run `pre-commit run --all-files` to fix the linting issues, and fix or disable the failing `test_capture_service` test.

aloisklink (approved):
Great!

PR #296, opened 2022-09-26
Remove deprecated `tempnam()` from eloop tests

aloisklink:
Removes the usage of `tempnam()`, which is deprecated and throws a warning on use.
The first line of the description in the tempnam manpage says:

> _Never use this function_.  Use mkstemp(3) or tmpfile(3) instead.

I've made a new file called `tests/utils/tmpdir.h` that contains the following functions, that can be used as CMocka setup/teardown functions.

- `int setup_tmpdir(void **test_state);`
- `int teardown_tmpdir(void **test_state);`

### DIFF
--- a/tests/utils/tmpdir.c
+++ b/tests/utils/tmpdir.c
@@ -1,0 +1,67 @@
+#define _XOPEN_SOURCE 700 /* mkdtemp() is part of POSIX 700 spec*/
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+#include <stdint.h>
+#include <cmocka.h>
+// used to delete directory recursively
+#include <ftw.h>
+
+#include <stdlib.h>
+
+#include "utils/log.h"
+
+#include "tmpdir.h"
+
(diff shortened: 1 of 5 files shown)


tests/utils/tmpdir.c
```diff
@@ -0,0 +1,67 @@
...
+#include <ftw.h>
+
+#include <stdlib.h>
+
+#include "utils/log.h"
+
+#include "tmpdir.h"
+
+// recursively delete folder using ftw
+static int rm_file(const char *pathname, const struct stat *sbuf, int type,
```
mereacre, line 17: Do you think it makes sense to add these nice files to `os.h`?

mereacre (approved):

tests/utils/tmpdir.c
aloisklink: Ehhhh, maybe. The problem is, if we add it to `os.h`, we'd also want to test every branch of it.

Basically, if it's not used in `src/`, we should probably keep the functions in `tests/`.

PR #297, opened 2022-09-26
Added const qualifiers

mereacre:
Nothing special, just `const` :)


aloisklink (approved):
Looks good to me!

Ideally we'd use `const char* const argv[]`, but that will probably be a bit more work to refactor!

PR #298, opened 2022-09-26
refactor(log): delete `ename.h` header

aloisklink:
The C11 standard already includes a function that does this for us, see https://en.cppreference.com/w/c/string/byte/strerror

`char* strerror( int errnum );` returns the string description of an error number. It's not threadsafe, but this function is only called between `lock()` and `unlock()` functions, so this is fine.

In C11, there's also the threadsafe `errno_t strerror_s( char *buf, rsize_t bufsz, errno_t errnum );` but it's slightly slower and more difficult to use.


mereacre (approved):

mereacre:
As usual:
```
2022-09-26 20:38:46.195  ERROR sqlite_header.c:614: execute_sqlite_query fail: CREATE TABLE IF NOT EXISTS eth (timestamp INTEGER NOT NULL, id TEXT NOT NULL, caplen INTEGER, length INTEGER, ifname TEXT, ether_dhost TEXT, ether_shost TEXT, ether_type INTEGER, PRIMARY KEY (timestamp, id));
[  ERROR   ] --- "/home/runner/work/edgesec/edgesec/tests/capture/middlewares/test_sqlite_header.c:51: init_sqlite_header_db: disk I/O error" != ""
[   LINE   ] --- /home/runner/work/edgesec/edgesec/tests/capture/middlewares/test_sqlite_header.c:78: error: Failure!
```

I think we need at what is the problem with sqlite.

aloisklink:
> I think we need at what is the problem with sqlite.
It's reported under https://github.com/nqminds/edgesec/issues/252

Personally, my gut feeling is that it's an issue with SQLite and the GitHub Actions runner file-system. Maybe wrapping the tests in a would work:

```c
for(int retries = 0; retries < 3; retries++) {
  if(do_test() == ok) {
    goto success;
  }
}
assert_false("Reached max retries");
...
```

PR #301, opened 2022-09-26
build: support semver versioning for edgesec

aloisklink:
**Draft until #300 is merged**

Add support for versioning edgesec following [semantic versioning (semver)](https://semver.org/).

For example, this means we might have an edgesec alpha release called `0.1.0-alpha.2+sha.2d99d58c88e205e8436b155dd654f0da4f5864d7`.

Semver isn't yet supported by CMake, but there is a proposal to add it https://gitlab.kitware.com/cmake/cmake/-/issues/23649, and I've tried to copy the variable names.


mereacre (approved):
Looks good.

PR #305, opened 2022-10-03
Download `minIni` instead of embedding source in repo

aloisklink:
Download `minIni` from https://github.com/compuphase/minIni instead of keeping it in the `src/utils/` folder of this repo.

Benefits:
  - Less lines of code
  - Static linking to OS minIni for better security/performance
    (Ubuntu 20.04 contains the libminini-dev package)
  - Easier updates to minIni (switching to v1.3 will just be one line)
  - Better licensing terms (minIni has a static linking exception to their license)

I did a `diff --strip-trailing-cr /tmp/minIni/dev/minGlue.h ./minGlue.h` and found the exact commit https://github.com/compuphase/minIni/commit/b40dff4924461272f669814da7d0c9fdc8be6d94
that matches the `minIni.*` and `minGlue.h` files in our repo.

PR done while I was bug-fixing OpenWRT :cry:  (my laptop is sooo slow to compile)


mereacre (approved):
Great

PR #306, opened 2022-10-03
Fix string bug in run dhcp process

aloisklink:
Makes the parameters for `run_dhcp_process()` const. This means that the tests in `test_dnsmasq()` are now correct.

Additionally, this fixes a potential bug that occurred, since `basename()` may modify the input string and corrupt running the dhcp process.


mereacre (approved):
Good to find potential bugs.

aloisklink:
> Good to find potential bugs.

Yah, it's a bit annoying how `basename()` and `dirname()` may or may not modify their input strings.

And what is more confusing is that there is also a POSIX and a GNU `basename()`, and they both have slightly different behaviour (the POSIX version changes the input string, but the GNU version doesn't (but it also returns some strange results)).

PR #307, opened 2022-10-03
Refactor non-`const` string literals in `iptables`

aloisklink:
Modifying string literals is undefined behaviour in the C spec, and can cause all sorts of trouble (especially at higher optimisation levels).

A good way to prevent modifying string literals is to label them as `const`. This is the default behaviour in C++, but unfortunately, back in the early days of C, `const` didn't yet exist, and so we need to add an optional `-Wwrite-strings` compiler flag to enable making them `const` by default.

While waiting for OpenWRT to super slowly compile, I went through the `iptables` source files and made sure that we had no `-Wwrite-strings` warnings.

### DIFF
--- a/src/utils/iptables.c
+++ b/src/utils/iptables.c
@@ -299,8 +299,8 @@ int get_nat_rules(struct iptables_context *ctx) {
   return utarray_len(ctx->rule_list) ? 0 : -1;
 }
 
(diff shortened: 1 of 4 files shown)


mereacre (approved):
Great

PR #308, opened 2022-10-10
Fix string bug in hostapd

aloisklink:
Fix some string issues in `hostapd.c`.

The function `basename()` can modify it's input string, so we should make a copy of the input string before calling `basename()`. This also meant that I could make the entire input arg to `signal_ap_process` const.

I also did some refactoring of `get_hostapd_args()` to make all the string inputs `const`, as it's `undefined` behaviour to modify string literals in C (and there was a `basename()` call that might have modified it).


src/ap/hostapd.c
```diff
@@ -167,18 +167,18 @@ int generate_hostapd_conf(struct apconf *hconf, struct radius_conf *rconf) {
 }
 #endif
 
+void get_hostapd_args(const char *hostapd_bin_path,
+                      const char *hostapd_file_path,
+                      const char *hostapd_log_path,
+                      const char *argv[static 5]) {
```
aloisklink, line 173: In case you're wondering what this syntax is, see https://beej.us/guide/bgc/html/split/arrays-part-ii.html#static-for-arrays-in-parameter-lists

Essentially, you're guaranteeing to the compiler that `argv` will always have at **least** 5 elements. This is usually good from an auto-vectorization/SIMD standpoint.

PR #311, opened 2022-10-10
Make `hmap_str_keychar_get()` params const

aloisklink:
Set the params and return value of `hmap_str_keychar_get()` to `const`.

Also, there's no point in making the `hmap` param a pointer-to-a-pointer if we never modify the pointer, we might as well just pass a pointer and get a performance boost, simplify code, and make it more obvious that we're not changing anything.

Benefits:
  - Reduces the risk of modifying string literals (undefined behaviour by the C compiler)
  - Makes it unlikely for us to accidentally modify hash_maps using the get function.
  - Improves usage documentation
  - Performance increase, since we're moving from a `**` to a `*` (pointer to pointer to just a pointer) (the compiler might have optimized this with Link-time-optimization)

### DIFF
--- a/src/firewall/firewall_service.c
+++ b/src/firewall/firewall_service.c
@@ -153,7 +153,7 @@ struct fwctx *fw_init_context(hmap_if_conn *if_mapper,
     return NULL;
   }
 #else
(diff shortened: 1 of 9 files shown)


mereacre (approved):
Great

PR #312, opened 2022-10-10
Cleanup sockctl files

aloisklink:
Refactors the `sockctl` files.

This PR mainly just improves documentation, and adds `const` to as many params as I can (as it was breaking some other code that I was working on).

However, there is a small bug fix, performance fix, and some refactoring that I also did.

I'd recommend reviewing commit by commit to make things more manageable.

### DIFF
--- a/src/utils/sockctl.c
+++ b/src/utils/sockctl.c
@@ -465,7 +465,7 @@ int writeread_domain_data_str(char *socket_path, const char *write_str,
 
   char *trimmed_data = os_realloc(rec_data, strlen(rec_data) + 1);
   if (trimmed_data == NULL) {
-    log_errno("os_realloc failed to rello string %s", rec_data);
+    log_errno("os_realloc failed to relloc string %s", rec_data);
     goto cleanup_recdata;
   }
   // set to NULL so free(rec_data) does nothing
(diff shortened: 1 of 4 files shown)


src/utils/sockctl.c
```diff
@@ -463,13 +463,16 @@ int writeread_domain_data_str(char *socket_path, const char *write_str,
   // rtrim modifies the input string.
   (void)rtrim(rec_data, NULL);
 
-  *reply = os_strdup(rec_data);
-  if (*reply == NULL) {
-    log_errno("os_strdup failed to copy string %s", rec_data);
+  char *trimmed_data = os_realloc(rec_data, strlen(rec_data) + 1);
+  if (trimmed_data == NULL) {
+    log_errno("os_realloc failed to rello string %s", rec_data);
```
mereacre, line 468:
```suggestion
log_errno("os_realloc failed to realloc string %s", rec_data);
```

mereacre (approved):
Looks good

src/utils/sockctl.c
aloisklink: Good catch! I'll use the suggestion below, since that has the correct indentation.

```suggestion
    log_errno("os_realloc failed to relloc string %s", rec_data);
```

PR #313, opened 2022-10-10
Fix `ap_service` bugs

aloisklink:
Fixes some memory leak and buffer overflow bugs in the `ap_service` files
(as well as a bunch of cleanup to make params `const` when possible).

I'd recommend reviewing this PR commit by commit, since it's got quite a few changes in it.


mereacre (approved):
Looks good.

PR #315, opened 2022-10-11
refactor(squeue): make squeue params const

aloisklink:
**DRAFT UNTIL #265 IS MERGED**

(we may also need to refresh the PR so that all of the PR #265 commits disappear)

The current `src/utils/list.h` does not work with this PR, however newer versions of `list.h` contain commit https://w1.fi/cgit/hostap/commit/src/utils/list.h?id=2a7023ba6f4ef318f52efbd52c9f54459c06d6cb which fixes the `list.h` file.

---

Make params to the `squeue` functions `const` when possible.

I've also improved the documentation, to clarify what are input/output params, and when output params must have `free()`-ed.

### DIFF
--- a/src/utils/squeue.c
+++ b/src/utils/squeue.c
@@ -31,7 +31,7 @@ struct string_queue *init_string_queue(ssize_t max_length) {
   return queue;
 }
 
-int push_string_queue(struct string_queue *queue, char *str) {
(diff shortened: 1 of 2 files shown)


mereacre (approved):
Great.

PR #316, opened 2022-10-11
Test the IP_GENERIC_SERVICE using GitHub Actions

aloisklink:
Test the `IP_GENERIC_SERVICE` using GitHub Actions.

Currently, the only CMake preset that uses the `IP_GENERIC_SERVICE` is `recap`, so I've added a CTest preset for it too, so that we can run `IP_GENERIC_SERVICE` tests and to get coverage results.

Additionally, I've fixed the `const` errors in `ipgen` that were preventing `recap` from compiling.


mereacre (approved):
Looks good.

PR #317, opened 2022-10-11
Refactor `iface` functions so that they are `const`

aloisklink:
**Draft until #315 and #316 are merged**

This is because it relies on the new `list.h` function, as well as the new const `ipgen_` functions.

--

Makes the `iface_*` functions `const`, by making the following functions `const` too:
  - `ipgen_*` Will be added by PR #316 
  - `net_*` Added by PR #325
  - `uci_wrt_*`

Fixes quite a few `-Wwrite-string` error in `recap.c` file.


mereacre (approved):
Great

PR #318, opened 2022-10-11
Prevent modifying string literals in `runctl`

aloisklink:
Prevents modifying string literals in `runctl`, by making sure that all functions that use string literals are const (e.g. following `-Wwrite-strings`).

Additionally, I had to make a copy of string literals whenever we pass them to a function that might modify them (e.g. run_process() or basename()).

After this PR (and the other open ones), I managed to fully compile the `--preset openwrt` when `-Wwrite-strings` without any errors! This means we shouldn't have any undefined behaviour due to modifying strings in OpenWRT.


mereacre (approved):
Great

PR #319, opened 2022-10-13
Argv

mereacre:


src/capture/middlewares/header_middleware/dns_decoder.c
```diff
@@ -22,7 +22,7 @@
 void decode_dns_questions(uint8_t *payload, struct capture_packet *cpac) {
   uint16_t idx, i = 0, j = 0;
   for (idx = 0; idx < /*cpac->dnss.nqueries*/ 1; idx++) {
-    while ((((void *)&payload[i] - (void *)cpac->ethh)) < cpac->length) {
+    while ((uint32_t)(((void *)&payload[i] - (void *)cpac->ethh)) < cpac->length) {
```
aloisklink, line 25:
```suggestion
    while ((uint32_t)((&payload[i] - (uint8_t *)cpac->ethh)) < cpac->length) {
```

Doing pointer arithmetic on `void *` is not allowed under the C standard.
(However both GCC and Clang allow it as an extension).

Still, it's worth using `uint8_t *` instead, since it's more standard:
https://gcc.gnu.org/onlinedocs/gcc-10.1.0/gcc/Pointer-Arith.html#Pointer-Arith

src/utils/ipgen.c
```diff
@@ -50,12 +50,12 @@ void ipgen_free_context(struct ipgenctx *context) {
...
-int run_ip(char *path, char *const argv[]) {
+int run_ip(char *path, const char *const argv[]) {
   return run_argv_command(path, argv, NULL, NULL);
 }
 
 int ipgen_new_interface(char *path, char *ifname, char *type) {
-  char *argv[7] = {"link", "add", "name", ifname, "type", type, NULL};
+  const char *argv[7] = {"link", "add", "name", ifname, "type", type, NULL};
   return run_ip(path, argv);
 }
```
aloisklink, line 60: I've already fixed all of these in https://github.com/nqminds/edgesec/commit/5599ac6f605651722634b1e08f4a7b69352f0b78, so we might get some merge conflicts here.

It might be easier to use a `git rebase -i main` and just delete https://github.com/nqminds/edgesec/pull/319/commits/0aed6030e7d5d94989fbe73f6803312893f80426, rather than have to worry about fixing all the merge conflicts.

aloisklink (approved):
Great! Thanks for fixing all those casts.

C's rules on implicit conversions are pretty confusing, especially when comparing signed/unsigned types.

aloisklink:
By the way, if you want a free T-shirt, you can mark this PR (and 3 other PRs you create this month on open-source repos) with `hacktoberfest-accepted` and sign up for https://hacktoberfest.com

src/capture/middlewares/header_middleware/dns_decoder.c
mereacre: Makes sense.

src/utils/ipgen.c
mereacre: I will resolve the conflicts.

PR #323, opened 2022-10-24
test(eloop): test handling NULL `eloop` param

aloisklink:
Test whether eloop functions correctly return -1 and log an error message if their `struct eloop_data *eloop` param is NULL.

Instead of duplicating the test 8 different times, I've used [X Macros](https://www.drdobbs.com/the-new-c-x-macros/184401387) to automate it.

My feeling is that it's cleaner than duplicating the code 8 times in 2 different places, even if it does use some crazy C pre-processor macros of macros.


mereacre (approved):
Very nice. Learned something today. Dr Dobbs articles are cool.

PR #324, opened 2022-10-24
Optimize the `hash` library

aloisklink:
Removes the `sha256_hash()` function from the `hash` library, since it's not being used.

I also removed unnecessary `#include`s, and converted `hash` into a CMake object library.
CMake `OBJECT` libraries are a bunch more efficient, since instead of creating `.a` or `.so` files, only an `.o` file is created.

See https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#object-libraries


src/utils/hash.c
```diff
@@ -23,15 +20,15 @@ uint32_t md_hash(const char *msg, size_t length) {
 
   // Loop over the message 32-bits at-a-time
   while (length >= 4) {
-    os_memcpy(&block, msg, sizeof(uint32_t));
+    memcpy(&block, msg, sizeof(uint32_t));
```
mereacre, line 23: Why `os_memcpy` renamed to `memcpy`?

aloisklink: It's to remove a dependency on `"allocs.h"`

It's just a macro to `memcpy` anyway, so I don't really see a point in using `os_memcpy`, see https://github.com/nqminds/edgesec/blob/84b412ff40720566c43134a2a23eb8e44c68aa50/src/utils/allocs.h#L98-L100

The only time I can see it being useful is if you're compiling to a platform that doesn't have a C standard library, but pretty much every platform has a C standard library.

mereacre (approved):

PR #326, opened 2022-10-25
test(uci_wrt): use temporary test data files

aloisklink:
The uci_wrt tests sometimes modify the `./tests/data/uci` files. This causes issues with `git switch`/`git rebase`.

This commit means we instead test on a copy of the test data in the build directory.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -6,7 +6,10 @@ include_directories (
 )
 
 if (USE_UCI_SERVICE)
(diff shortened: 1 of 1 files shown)


mereacre (approved):

aloisklink:
I'm deleting the `test/use-tmp-dir-to-test-uci-wrt` branch again, since it looks like the reason why this bug occurred in https://github.com/nqminds/edgesec/pull/339#discussion_r1016390637 was because that branch didn't originally contain this fix.

Now that PR #339 has been rebased to contain these changes, this bug no longer occurs.

PR #328, opened 2022-10-28
test(supervisor): fix segmentation fault

aloisklink:
Fix segmentation fault due to uninitialized `crypt_ctx`.

---

It might also be worth adding `if (crypto_ctx == NULL) {log_error("crypto_ctx == NULL"); return -1;}` to some functions.


mereacre (approved):
Thanks. You did it just as I was planning to do it. :)

PR #330, opened 2022-10-28
Use `cmake` to configure protobuf-c

### LINKED ISSUE #333: Failing CI builds due to automake version mismatch
**Describe the bug**

GitHub Actions CI occasionally fails with:

```
configure.ac:13: error: version mismatch.  This is Automake 1.16.1,
configure.ac:13: but the definition used by this AM_INIT_AUTOMAKE
configure.ac:13: comes from Automake 1.16.5.  You should recreate
configure.ac:13: aclocal.m4 with aclocal and run automake again.
configure.ac:13: warning: user variable 'ETAGS' defined here ...
```

The issue seems to be with GitHub Actions runners, some of their runners seem to be using an older `automake` version. Restarting the CI action usually means that the CI job will be rerun on a working runner.

aloisklink:
Use `cmake` to configure `protobuf-c` instead of `./configure`.

The `./configure` script that comes with protobuf-c is too simple to understand how to configure things for the Morello architecture.

~Currently only a draft, since for some reason, the `debian` cross-compile isn't working properly.~ Working! `libprotobuf-c` was still being built as x86_64, even when cross-compiling to arm64, but I copied the config from the `lib/uci.cmake` files and that seems to have fixed it.

Should fix #333, since CMake should be less buggy than a `./configure` script.


mereacre (approved):
Great

aloisklink:
> Great

Btw, I've stuck this in the `CheriBSD Support` milestone, so that we can easily track what PRs/Issues were needed to get CheriBSD support working.

PR #331, opened 2022-10-31
Allow using existing `libprotobuf-c` library

aloisklink:
Gets `BUILD_PROTOBUFC_LIB=OFF` working.

This means that we can build the `.deb` without cross-compiling `libprotobuf-c`, and instead dynamically link to an existing `libprotobuf-c` (e.g. the [libprotobuf-c1](https://packages.ubuntu.com/jammy/libprotobuf-c1) package).

As a side-note, I've added `EXCLUDE_FROM_ALL` to all protobuf targets, so they're only compiled:

  - a) if `USE_CAPTURE_SERVICE` is true (e.g. build `recap` executable) (disabling this is currently broken) AND/OR
  - b) if `USE_PROTOBUF_SERVICE` is true (adds the protobuf service to edgesec)


mereacre (approved):
Great

PR #332, opened 2022-10-31
Fix failing `cmocka` build, due to upstream bug

aloisklink:
Building CMocka was broken due to https://gitlab.com/cmocka/cmocka/-/issues/50.
Because of this, building CMocka was failing on FreeBSD/CheriBSD.

This commit points to my fix in https://gitlab.com/cmocka/cmocka/-/merge_requests/51

Additionally, I've also switched the download from using `git` to using plain `https://`. This is much faster, and means that we don't need to worry about installing `git` on FreeBSD.

It also means we can now delete the `build/linux` folder without using `rm -rf`, since no files are protected. (OpenWRT builds still need to use `git`, so it doesn't work everywhere).

Finally, in order to actually test the CMocka build, I've made it so that the `cmake --preset clang` test actually builds CMocka. That way, if we do ever accidentally break CMocka again, at least we'll know straight away.


mereacre (approved):
Great

PR #334, opened 2022-11-01
Use `AddCMockaTest.cmake` script to simplify test scripts

aloisklink:
## Summary

Uses the `AddCMockaTest.cmake` script from the CMocka library to simplify building and adding tests.

Because it's only a build script (it's not a statically-linked executable), we don't need to add licensing information to the `edgesec` software under the BSD 3-Clause license.

#### [build(cmocka): add AddCMockaTest.cmake script](https://github.com/nqminds/edgesec/commit/23da1a8b1344b1d33ef2d4063475c560a10c7ab6) 

Adds the `AddCMockaTest.cmake` helper script from https://gitlab.com/cmocka/cmocka/-/blob/a7a1ae83d567ccd6b469e157fc0a3db872039a49/cmake/Modules/AddCMockaTest.cmake

#### [docs(AddCMockaTest): add licensing info](https://github.com/nqminds/edgesec/commit/a8319cda47b6fa5ec057670be8b10d1eff23a71e) 

Add licensing and copyright information for AddCMockaTest.cmake

It uses the BSD 3-Clause license.

#### [test: use AddCMockaTest to simplify cmake files](https://github.com/nqminds/edgesec/commit/5cd467ed6ea2c0c6a774e09ade8bef3b68585402) 

Use the AddCMockaTest.cmake script to simplify the test CMake build scripts.

By default, all `LINK_LIBRARIES` are linked privately by default, since the tests are executables and are not expected to be linked to by other CMake targets.

There is a LINK_OPTIONS variable that the script uses, but confusingly,
it is not actually passed to CMake's `LINK_OPTIONS`. It's instead passed to [`LINK_FLAGS`][1]

[1]: https://cmake.org/cmake/help/latest/prop_tgt/LINK_FLAGS.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/lib")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/docs")
 
-include(CTest)
+include(AddCMockaTest)
 include(FetchContent)
 include(ExternalProject)
 include(CMakeDependentOption)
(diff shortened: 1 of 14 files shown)


mereacre (approved):
Very nice.

aloisklink:
> Very nice.

Huh, this PR actually fixed a bug. Before this PR, we were never actually testing `test_iface_mapper`. Proof that refactoring and deleting code is useful :)

https://github.com/nqminds/edgesec/blob/5cd467ed6ea2c0c6a774e09ade8bef3b68585402/tests/utils/CMakeLists.txt#L14-L16

PR #335, opened 2022-11-01
Add support for compiling edgesec using Ninja

aloisklink:
Ninja is much faster than GNU Make, and is now the default CMake generator for OpenWRT, see https://github.com/openwrt/openwrt/commit/0c7c24d40aedcee25d5243e25a87d38246be128b.

This PR adds support for compiling edgesec using Ninja.

### [build: support non-makefile CMake generators](https://github.com/nqminds/edgesec/commit/ad71f799b5165503890726bbe6331956c469d715)

Currently, all of our `ExternalProjects` using `make` are using `$(MAKE)` so that their builds are run in the same `make` invocation as the main repo.

(e.g. this means things like multiprocessing get shared with all ExternalProjects).

However, this does not work when the main project is using a non-Makefile generator.

Instead, I've changed the CMake files to use `make` when not running in a Makefile generator.

Unfortunately, this change actually means that build edgesec with Ninja is **slower** than using GNU Make. This is because then the ExternalProjects have to run in individual GNU Make sessions, which I've set to run in single-process mode, while normally, they can use the shared GNU Make job server (interestingly, both CMake and OpenWRT use a fork of Ninja that does have compatibility with GNU Make job server, but it hasn't yet been upstreamed, see https://github.com/ninja-build/ninja/issues/1139)

#### Benchmarks

GNU Make

```console
alois@nqm-alois-entroware:~/Documents/edgesec (build/add-ninja-support)$ export PRESET=linux; cmake --preset "$PRESET" && time nice -n19 cmake --build --preset "$PRESET" -j=$(nproc)
...
[100%] Built target edgesec

real	0m39.455s
user	1m58.707s
sys	0m12.376s

```

Ninja

```console
alois@nqm-alois-entroware:~/Documents/edgesec (build/add-ninja-support)$ export PRESET=linux; cmake --preset "$PRESET" -G Ninja && time nice -n19 cmake --build --preset "$PRESET" -j=$(nproc)
...
[333/333] Linking C executable src/edgesec

real	1m4.380s
user	1m37.235s
sys	0m9.267s
```

### [build: support using Ninja with CMake](https://github.com/nqminds/edgesec/commit/c073ac8a78c1754df3ebe3972daf265b15e80d2d)

One of the reasons why Ninja is so fast is that underspecified build dependencies are not supported. However, these are used in edgesec.

To fix this, we need to manually declare the library files for our ExternalProject libraries, using `BUILD_BYPRODUCTS`.

Technically, all of the library files are actually generated by the ExternalProject INSTALL step, not the BUILD step, but we only need to add these rules to make Ninja happy, so it's fine.

In the future, if CMake adds an `INSTALL_BYPRODUCTS` option, we should use that instead.

See https://gitlab.kitware.com/cmake/cmake/-/issues/23056

### [ci(recap): build recap using Ninja](https://github.com/nqminds/edgesec/commit/6bd9c372bd8ea2860d5e6308e9e8e0dfd890b60e) 

Builds the Recap library using Ninja in GitHub Actions as a test case for Ninja compatibility.

This is may be slightly slower than using GNU Make, since then the ExternalProjects have to run in individual GNU Make sessions (see my comment earlier in this PR). But GitHub Actions only has 2 cores by default, so this shouldn't make too much of a difference.

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -148,10 +148,13 @@ jobs:
       - name: Install Dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install devscripts equivs -y # install mk-build-depends
+          sudo apt-get install devscripts equivs ninja-build -y # install mk-build-depends
           sudo mk-build-deps --install --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes'  debian/control
-      - name: Configure
-        run: cmake --preset recap -DCMAKE_BUILD_TYPE=Release
+      - name: Configure using Ninja
+        # Ninja is actually slightly slower than using Make, since then
+        # the ExternalProjects have to run in invidividual GNU Make sessions
+        # It's mainly here just to confirm that a Ninja build works.
+        run: cmake --preset recap -DCMAKE_BUILD_TYPE=Release -G Ninja
       - name: Build
         run: cmake --build --preset recap -j="$(($(nproc) + 1))"
       - name: Archive built recap
--- a/lib/mnl.cmake
+++ b/lib/mnl.cmake
@@ -3,6 +3,13 @@ if (BUILD_ONLY_DOCS OR NOT USE_NETLINK_SERVICE)
 elseif(BUILD_MNL_LIB)
   set(LIBMNL_INSTALL_ROOT "${CMAKE_CURRENT_BINARY_DIR}/lib/mnl")
 
+  if ("${CMAKE_GENERATOR}" MATCHES "Makefiles")
+    set(MAKE_COMMAND "$(MAKE)") # recursive make (uses the same make as the main project)
+  else()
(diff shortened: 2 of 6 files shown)


mereacre (approved):
Cool. Looking forward to the improvment.

aloisklink:
> Cool. Looking forward to the improvment.

It's slower for the first run (since all of the `ExternalProject` calls still use `make`, which doesn't combine with the root `Ninja` run), but the next runs should all be slightly faster.

Just remember to use `-G Ninja` to ask CMake to build `.ninja` files instead of `Makefiles`, see https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#ninja-generators

This PR is mainly here because the OpenWRT SDK uses Ninja for CMake builds, and their Ninja is a fork that is much faster, since it's compatible with GNU Make's multiprocessing job server.

PR #338, opened 2022-11-04
Download ExternalProject files into `EP_DOWNLOAD_DIR`

aloisklink:
Add the CMake `EP_DOWNLOAD_DIR` cache variable.

This allows for setting a constant ExternalProject download directory.

This is useful for two reasons:
  - A single download directory can be shared for all build presets
  - Allows building/configuring when there is no internet connection
    (e.g. OpenWRT SDK or default `pdebuild` environment)

This is what the download directory may look like:

```console
alois@me:~/edgesec (build/download-uthash-at-configure-time)$ ls build/dl/
cmocka-c911f1cb11479435818c59fd265c103b4d8c66fb.tar.bz2  openwrt-sdk-19.07.10-mvebu-cortexa9_gcc-7.5.0_musl_eabi.Linux-x86_64.tar.xz
hostapd-2.10.tar.gz                                      sqlite-autoconf-3310100.tar.gz
libmnl-1.0.4.tar.bz2                                     uthash-2.1.0.tar.gz
libpcap-1.10.1.tar.gz                                    util_linux-2.37.2.tar.gz
```

The goal is to use the OpenWRT SDK `dl` folder when building with the OpenWRT SDK folder, which would allow us to fix the OpenWRT download issue.


mereacre (approved):
Great

PR #340, opened 2022-11-07
refactor(squeue): cleanup squeue includes

aloisklink:
`squeue.h` was including `squeue.h` (aka it was including itself!), which was causing a circular dependency error in a static analysis software.

I've also removed the unneeded includes that it doesn't need, which should hopefully speed up builds slightly.


mereacre (approved):
Great

PR #341, opened 2022-11-07
Support compiling edgesec using standard ISO C11

aloisklink:
Support compiling edgesec using standard ISO C11 (e.g. `-std=c11` flags).

When we do use non-standard C features, we enable the [Feature test macro](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html) we need.

---

I'd recommend reviewing this PR commit by commit, since it's pretty big :)


src/utils/list.h
```diff
@@ -56,9 +56,9 @@ static inline int dl_list_empty(struct dl_list *list) {
   return list->next == list;
 }
 
-static inline ssize_t dl_list_len(struct dl_list *list) {
+static inline unsigned int dl_list_len(struct dl_list *list) {
   struct dl_list *item;
-  ssize_t count = 0;
+  int count = 0;
```
mereacre, line 61:
```suggestion
  unsigned int count = 0;
```

mereacre (approved):
Great work. A lot of compiler definitions. Will you be able to write a small doc for reference of which definitions and where it is used in the code (maybe as a summary for me to know)? If it is too much work then don't bother.

src/utils/list.h
aloisklink: Actually, this is on purpose!

It's taken directly from the hostapd source code: https://w1.fi/cgit/hostap/tree/src/utils/list.h?id=0ccb3b6cf25e0c432de092a5298be96accd18c68#n54

I'm not 100% sure why, maybe it's because in C, signed ints overflowing is undefined behaviour, so the compiler can assume that the loop will never run forever?

If it's important, we could have a look at the `hostapd` git history, and see if they explain why, since if they don't mention it, we could send a patch to hostapd.

aloisklink:
@mereacre, this isn't actually part of CheriBSD support, it's part of the OpenWRT 21.02 support.

A bunch of these types are actually standard BSD functions, hence why we need to declare `_BSD_SOURCE` to use these :)

> Great work. A lot of compiler definitions. Will you be able to write a small doc for reference of which definitions and where it is used in the code (maybe as a summary for me to know)? If it is too much work then don't bother.

https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html is probably the best documentation for these. Basically, whenever we use a non-standard C function (after this PR, GCC should shout at you), we may need to declare a header guard.

Most of the man pages for your custom function would explain them. E.g., the Linux man page for `strdup` (https://man7.org/linux/man-pages/man3/strdup.3.html) says:

>    Feature Test Macro Requirements for glibc (see [feature_test_macros(7)](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html)):
> 
>        strdup():
>            _XOPEN_SOURCE >= 500
>                || /* Since glibc 2.12: */ _POSIX_C_SOURCE >= 200809L
>                || /* Glibc <= 2.19: */ _BSD_SOURCE || _SVID_SOURCE
> 
>        strndup():
>            Since glibc 2.10:
>                _POSIX_C_SOURCE >= 200809L
>            Before glibc 2.10:
>                _GNU_SOURCE
> 
>        strdupa(), strndupa():
>            _GNU_SOURCE

src/utils/list.h
aloisklink: :+1: 

Once #265 is merged, this wouldn't matter anyway, since we'd be using the hostapd code directly then.

PR #344, opened 2022-11-09
fix: segfault on add uwrt  properties

mereacre:
Fixes a segmentation fault in uwrt.


aloisklink (commented):
Looks good!

Don't merge, I'll back-port this commit so that it's closer to the `v0.1.0-alpha.0` release.

By the way, it might be worth moving all of the `properties` stuff into a different scope so that we don't get a segmentation fault.

https://github.com/nqminds/edgesec/pull/344/files#diff-8faadcc457b05e2b5d4ea280c5e2039bdaac7a2fa380d202ce839b30ec322cc1R894-R966

aloisklink (approved):
I've `git rebase`-d the commits onto c7e67ac263df7f26215beaed44b14f4dc84d78c3

PR #345, opened 2022-11-11
Update create-debs.yml

mereacre:
Commented the non-code coverage tests, so that only the code coverage tests run.

This is because PR #339 seems to have some issues with running GitHub Actions twice (tests are killed by `SIGHUP`).


aloisklink (changes requested):
Ah, now I kind of remember why we duplicated this.

I think  the actual way to fix this PR would be to just comment out this line and nothing else (indentation of the `#` might matter) https://github.com/nqminds/edgesec/blob/0925c72a5f7b05decf632f126647828230c2ad27/.github/workflows/create-debs.yml#L69 and nothing else.

It's probably worth also putting a comment saying: `# temporarily disabled due to GitHub Actions SIGHUP issue`

aloisklink (approved):

PR #346, opened 2022-11-11
build: enable `-Wwrite-strings` for the `src/` dir

aloisklink:
In C, it's undefined behaviour to modify a string literal.

However, because C previously didn't have a `const` specifier, for backwards compatibility, string literals have the type `char *` not `const char *` by default.

The GNU `-Wwrite-strings` option changes string literals to have type `const char *` by default, which should make it less likely for us to modify string literals accidentally.

This is currently only enabled in the `src/` dir since it fails in the `test/` dir.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,9 @@
-# Error on all C/C++ warnings in the src/ directory if making a Debug build
-add_compile_options($<$<CONFIG:Debug>:-Werror>)
+add_compile_options(
+  # Error on all C/C++ warnings in the src/ directory if making a Debug build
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Good point. We can keep it disabled for tests folder as it doesn't matter for the correcteness.

aloisklink:
> Good point. We can keep it disabled for tests folder as it doesn't matter for the correcteness.

:+1: Because we compile tests with no optimizations, it's unlikely that the compiler will break anything.

Still, it's worth enabling in the `tests/` folder, just because it might catch functions we should be making `const`.

PR #347, opened 2022-11-14
feat: Added pragma options to speedup the recap

mereacre:
Added the below sqlite pragma options to speed the db recapture:
```
  PRAGMA journal_mode = OFF;
  PRAGMA synchronous = 0;
  PRAGMA cache_size = -4000;
  PRAGMA locking_mode = EXCLUSIVE;
  PRAGMA temp_store = MEMORY;
```

### Performance difference:

Seems to be about 112× faster!!!

> The speed difference is huge 16 secs compared to around 30 minute for the 400 Mb dataset.
>
> From https://github.com/nqminds/edgesec/pull/347#issuecomment-1313823525

### DIFF
--- a/src/recap.c
+++ b/src/recap.c
@@ -72,12 +72,9 @@ struct pcap_stream_context {
 };
 
 const char *pragma_options[] = {
-  "PRAGMA journal_mode = OFF",
-  "PRAGMA synchronous = 0",
-  "PRAGMA cache_size = -4000",
-  "PRAGMA locking_mode = EXCLUSIVE",
-  "PRAGMA temp_store = MEMORY"
-};
+    "PRAGMA journal_mode = OFF", "PRAGMA synchronous = 0",
+    "PRAGMA cache_size = -4000", "PRAGMA locking_mode = EXCLUSIVE",
+    "PRAGMA temp_store = MEMORY"};
 
 void show_app_version(void) {
   fprintf(stdout, "recap app version %s\n", EDGESEC_VERSION);
--- a/src/recap.c
+++ b/src/recap.c
@@ -25,13 +25,12 @@
 
 #include "version.h"
 #include "utils/os.h"
+#include "utils/sqliteu.h"
 #include "capture/middlewares/header_middleware/sqlite_header.h"
 #include "capture/middlewares/header_middleware/packet_decoder.h"
 #include "capture/middlewares/header_middleware/packet_queue.h"
 #include "capture/middlewares/protobuf_middleware/protobuf_middleware.h"
 
-#define PCAP_READ_INTERVAL 10 // in ms
-#define PCAP_READ_SIZE 1024   // bytes
 #define IFNAME_DEFAULT "ifname"
 
 #define OPT_STRING ":p:f:i:kdhv"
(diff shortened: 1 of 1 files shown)


aloisklink:
Hi @mereacre, can you add some details on the actual speed difference of the `PRAGMA` options?

E.g. running `pv input.pcap | ...` before and afterwards and documenting the speed difference.

My gut feeling is that we can probably greatly speed things up by disabling `auto-commit`, e.g. by doing the following:

```sql
BEGIN TRANSACTION;
INSERT ...;
INSERT ...;
INSERT ...;
INSERT ...;
END TRANSACTION;
```

It should also be a lot safer, since the above PRAGMA options are pretty dangerous, since they might corrupt the `connection.sqlite` db.

mereacre:
I would first the current settings on the 4GB dataset. Then I can add more options or add the commit functionality. The speed difference is huge 16 secs compared to around 30 minute for the 400 Mb dataset.

mereacre:
These options are only for the recap tool. It has no connection to edgesec. Unless we want to save directly into the edgesec capture db, which I wouldn't advice.

We definitely should test the commit functionality on the actual edgesec capture code, i.e., header middleware.

aloisklink (changes requested):
> Unless we want to save directly into the edgesec capture db

That's the issue I have. Currently, the manysecured-app's PCAP upload function does exactly that, it uploads to the same capture db as the one that edgesec creates :(

So it's currently unsafe to use these PRAGMAs, unless we remove the PCAP upload functionality from the app (maybe we should hide them behind an `--unsafe` CLI option too, just to be on the safe side).

>  We should first test the current settings on the 4GB dataset (I can do it myself if you send me the link to the file). Then I can add more options or add the commit functionality. The speed difference is huge 16 secs compared to around 30 minute for the 400 Mb dataset.

Awesome!!! That's a big difference.

I don't have the PCAP file on my laptop. I'm unfortunately in the Nquiringminds' office today, and that means my internet/laptop is super slow (I've been trying to download a 4 GiB PCAP since this morning and it still has 80 minutes to go).

You could try downloading it from Google Drive: https://drive.google.com/drive/folders/1GdGmlikrLQRhfS0tPzVo1Ztw_m2FZKu1?usp=share_link


### Changes Requested

Either:

a) hide these unsafe `PRAGMA`s behind a `--unsafe` CLI option (and warn that they may corrupt your `capture.sqlite` database)
b) use safe `PRAGMA`s and other performance optimizations (e.g. SQLITE manual commits)

aloisklink:
> The best option is I think if I add a new command line argument were the user can append the pragma options as desired.

Yep, that would also work! As long as the default `PRAGMA` options are safe, I'll be happy with it.

aloisklink:
> We should first test the current settings on the 4GB dataset

Tested, running the import on 5.5 GB PCAP file on main (bf10bcf81703e24f524bec93f527d6cca90b6b8d) had an ETA of more than 6 days on my laptop.

Using this branch (269923e1bca3ce3b3d9ff0389545cde2e04f4e41) turned this ETA into 10 minutes.

Please be aware that this is using the `linux` preset, which I believe has no optimizations on by default, that might be one of the reasons slowing things down.

mereacre:
@aloisklink I prefer your PR #348 so I will close this one.

PR #348, opened 2022-11-21
Use SQLITE transactions when inserting PCAP data in `recap`.

aloisklink:
Modifies the recap executable to pipe all of the PCAP data into the sqlite database using a single SQLITE transaction.

This has two benefits:

1. On an error, the entire transaction is rolled-back, so we won't have a half imported recap file.
2. Running recap is much, much, much, much faster.
   **A quick local test went from 12.0KiB/s to 40.4 MiB/s, which is a ~3300x increase in speed!**

### Benchmarks

When using the `cmake --preset linux` setting:

#### Before (commit 12a7aa444da23f8225288aafe67dda0696f5d741)

After running for 5 minutes, it has only imported 2 MiB of the PCAP file.

```console
alois@laptop:~/Documents/edgesec ((12a7aa44...))$ timeout 5m bash -c 'pv ~/Downloads/2021_11_24_Active.pcap | ./build/linux/src/recap -f /tmp/recap-pcap-benchmark.sqlite'
Opened db at /tmp/recap-pcap-benchmark.sqlite
1.98MiB 0:04:59 [12.0KiB/s] [>                              ]  0% ETA 9:19:41:17
```

#### After (commit ce3c5f5b93f27ac864997f8f8cdc243e0caa23c8)

After running for 2:18 minutes, it had imported 5.48 GiB of the PCAP file (i.e. all of it).

Interestingly, it looks like it's now bottlenecked by CPU speed.

```console
alois@laptop:~/Documents/edgesec (optimize-recap)$ timeout 5m bash -c 'pv ~/Downloads/2021_11_24_Active.pcap | ./build/linux/src/recap -f /tmp/recap-pcap-benchmark.sqlite'
Opened db at /tmp/recap-pcap-benchmark.sqlite
5.48GiB 0:02:18 [40.4MiB/s] [================================>] 100%            
Processed pcap size = 5883456163 bytes
Processed packets = 18555332
```

I also tried compiling with `-DCMAKE_BUILD_TYPE=Release` (e.g. build with optimizations) and it actually seemed to slow things down to 2:40 minutes, but maybe that's just because my laptop has a terrible quality heatsink and it was fully saturated :laughing:

### Notes

I've also done a bit of minor refactoring while I was in the function, so please review this PR commit by commit!

It's up to you if this closes https://github.com/nqminds/edgesec/pull/347.
The optimizations in that PR should be cumulative with this PR, but it is at the cost of adding additional danger.

PR #349, opened 2022-11-21
Continuous pcap stream integration

mereacre:
Added libpcap streaming options.

Capture packets and svaes to SQLite DB or pipes to named pipes using protobuf serialization.

To Do:
 - Need to to add a queue or ring buffer to store arriving packets as there's a high drop rate.


src/recap.c
```diff
@@ -85,6 +88,7 @@ void show_app_help(char *app_name) {
   fprintf(stdout, "\t-p filename\t Path to the pcap file name\n");
   fprintf(stdout, "\t-f filename\t Path to the capture db or pipe\n");
   fprintf(stdout, "\t-i interface\t Interface name to save to db\n");
+  fprintf(stdout, "\t-t\t\t Capture from interface using libpcap\n");
```
aloisklink, line 91: Why `-t` :tea: ?

I'm mainly asking because in PR #348, I used the exact same character (`-t` for `transaction`) (great minds think alike), see https://github.com/nqminds/edgesec/blob/736e0917212ffc27e4a9e5fa25d03d9a22934102/src/recap.c#L122-L132 :)

I did try looking at `getopts_long(` so that I could use `--transaction`, but that's a glibc specific extension, so we may struggle to get it working properly on FreeBSD.

src/recap.c
```diff
@@ -390,11 +411,70 @@ int process_pcap_stream_state(struct pcap_stream_context *pctx) {
   }
 }
 
+int process_pcap_stream(char *pcap_path, struct pcap_stream_context *pctx) {
```
aloisklink, line 414:
```suggestion
int process_pcap_stream(const char *pcap_path, struct pcap_stream_context *pctx) {
```

src/recap.c
```diff
@@ -390,11 +411,70 @@ int process_pcap_stream_state(struct pcap_stream_context *pctx) {
...
+  if (pcap_path != NULL) {
+    if ((pctx->pcap_fd = fopen(pcap_path, "rb")) == NULL) {
+      log_errno("fopen");
```
aloisklink, line 417: It's always worth being a bit more verbose, especially for errors that are likely to come from user error (e.g. me accidentally writing `recap -p my-file.pcp` instead of `.pcap` or something)!

```suggestion
      log_errno("fopen failed for pcap file %s", pcap_path);
```

aloisklink (commented):
Not a full review, but:

> Need to to add a queue or ring buffer to store arriving packets as there's a high drop rate.

My gut feeling is that you should also wait until the buffer is full, then do a `BEGIN TRANSACTION`, insert the entire buffer, then a `COMMIT TRANSACTION`. It will greatly increase the throughput.

---

You may also want to add my changes from PR #348 in too, since I've done a bit of refactoring, and I've added some tests for `recap`.

---

One last thing, can you think of a better name for this PR! Maybe something like `recap from libpcap stream`, or `` Add libpcap stream input to `recap` ``.

src/recap.c
mereacre: It was a random selection. I can change -t to something else.

mereacre:
There's still works to be done. Because at the moment the libcap is dropping packets when they are not precessed in time. So I will have to use the queue in the same way as I used for the header middleware.

src/recap.c
aloisklink: > It was a random selection

I think being vegan may have given you psychic superpowers, so you read my mind and realised predicted which option char I was using :wink: :brain:

aloisklink:
Hey @mereacre, my gut feeling is to undo https://github.com/nqminds/edgesec/pull/349/commits/bf38277ef865a3663f6a30f4ea46da067ac5d515, but instead change: https://github.com/nqminds/edgesec/blob/2582047c213c9e71cf3e170182c4e0bd922d38c5/src/recap.c#L566 to

```c
  if (sqlite_transaction_type == DEFAULT) {
    if (capture == 0) {
      sqlite_transaction_type = SINGLE_TRANSACTION;
    } else
      // disable transactions, since `recap` might run forever
      sqlite_transaction_type = DISABLED;
    }
  }
```

The way `cat my-pcap-file.pcap | recap` and `recap -p my-pcap-file.pcap` will still be super fast by default, but when you use `-n` transactions will be disabled by default.

In the future, it might be worth making an `sqlite_transaction_type = WHEN_BUFFER_FULL` too, since I'm sure that will greatly increase your insertion speed for this PR.

aloisklink:
> I simplified the options. So now there's only one flag "-t" meaning with commit. More advanced options can be added but we need to see if there's a use for them.

It does mean that it's now disabled by default though, which I don't really like.

Ideally it would enabled when importing a `.pcap` file, and disabled by default when using a continuous stream from `libpcap`, but I guess it's not too hard to just add the `-t` flag.

mereacre:
I agree. I will enable by default for pcap capture.

If you prefer I can enable by default the commit for pcap file and stream.

src/capture/middlewares/header_middleware/packet_decoder.c
```diff
@@ -363,6 +363,10 @@ int extract_packets(const char *ltype, const struct pcap_pkthdr *header,
   cpac.length = header->len;
 
   os_strlcpy(cpac.ifname, interface, IF_NAMESIZE);
+
+  char id[MAX_RANDOM_UUID_LEN];
+  generate_radom_uuid(id);
+
   os_strlcpy(cpac.id, id, MAX_RANDOM_UUID_LEN);
```
aloisklink, line 370: We can simplify to the following:

```suggestion
  generate_radom_uuid(cpac.id);
```

src/capture/middlewares/protobuf_middleware/protobuf_middleware.h
```diff
@@ -13,13 +13,24 @@
 
 #include "../../middleware.h"
 
+/**
+ * @brief pipe the serialised protobuf tuple packets
+ *
+ * @param path[in] The pipe file path
+ * @param fd[in] The pipe file descriptor
```
aloisklink, line 20: This variable may be changed by the function.

```suggestion
 * @param fd[in,out] The pipe file descriptor.
 * If `0`, this will be set to a new fd pointing to @p path.
```

src/capture/pcap_service.h
```diff
@@ -105,6 +105,15 @@ int dump_file_pcap(struct pcap_context *ctx, char *file_path,
  */
 int inject_pcap(struct pcap_context *ctx, uint8_t *packet, size_t size);
 
+/**
+ * @brief Return pcap capture statistics
+ *
+ * @param ctx The pcap context
+ * @param ps The pcap_stat structure
```
aloisklink, line 112: Can you add `[in]` and `[out]` to these params? I don't know which is the input and which is the output.

(Adding `const` to the input param if possible will be nice too)

src/recap.c
```diff
@@ -23,63 +23,38 @@
...
 
+#include <eloop.h>
+
 #include "version.h"
 #include "utils/os.h"
 #include "capture/middlewares/header_middleware/sqlite_header.h"
 #include "capture/middlewares/header_middleware/packet_decoder.h"
 #include "capture/middlewares/header_middleware/packet_queue.h"
 #include "capture/middlewares/protobuf_middleware/protobuf_middleware.h"
+#include "capture/capture_service.h"
```
aloisklink, line 34: @mereacre, before you merge this, don't merge https://github.com/nqminds/edgesec/pull/116, otherwise it will cause merge conflicts (it's easy to autoformat that PR to fix the merge conflicts)

src/recap.c
```diff
@@ -146,20 +113,20 @@ void log_cmdline_error(const char *format, ...) {
...
   exit(EXIT_FAILURE);
 }
 
-void process_app_options(
-    int argc, char *argv[], uint8_t *verbosity, char **pcap_path,
-    char **out_path, char **ifname, int *pipe,
-    enum SQLITE_TRANSACTION_TYPE *sqlite_transaction_type) {
+void process_app_options(int argc, char *argv[], uint8_t *verbosity,
+                         char **pcap_path, char **out_path, char **ifname,
+                         int *pipe, int *capture, int *transaction) {
```
aloisklink, line 129: Is it worth using C99 boolean types instead?

See https://en.cppreference.com/w/c/types/boolean

src/recap.c
```diff
@@ -393,80 +395,241 @@ int process_pkt_read_state(struct pcap_stream_context *pctx) {
...
+      log_error("Failed to capture a lock on db %s, please retry this "
+                "command later",
+                pctx->out_path);
+    }
+
+    save_packets_from_queue(pctx);
+
+    if (execute_sqlite_query(pctx->db, "COMMIT TRANSACTION") < 0) {
+      log_error("Failed to commit packets to database %s", pctx->out_path);
+    }
```
aloisklink, line 552: Ooooh :) I like this syntax! Out of curiosity, do you know how much faster this makes your code?

src/recap.c
```diff
@@ -393,80 +395,241 @@ int process_pkt_read_state(struct pcap_stream_context *pctx) {
...
+}
+
+void eloop_tout_header_handler(void *eloop_ctx, void *user_ctx) {
+  struct recap_context *pctx = (struct recap_context *)user_ctx;
+
+  if (is_packet_queue_empty(pctx->pq) < 1) {
+    log_trace("Commiting packets to %s database", pctx->out_path);
+    if (execute_sqlite_query(pctx->db, "BEGIN IMMEDIATE TRANSACTION") < 0) {
+      log_error("Failed to capture a lock on db %s, please retry this "
+                "command later",
```
aloisklink, line 544: You may need to improve this error. I'd even drop down this to a `log_warning()`, because I'd expect this to happen pretty often due to `SQLITE_BUSY`.

src/recap.c
```diff
@@ -393,80 +395,241 @@ int process_pkt_read_state(struct pcap_stream_context *pctx) {
...
+int process_pcap_capture(struct recap_context *pctx) {
+  struct eloop_data *eloop = NULL;
+  int exit_code = -1;
+
+  if ((eloop = eloop_init()) == NULL) {
+    log_error("eloop_init fail");
+    goto process_pcap_capture_fail;
+  }
+
+  struct pcap_context *pc = NULL;
```
aloisklink, line 572: I'm surprised GCC didn't throw an error for this.

Fixes:

> ```
> /home/runner/work/edgesec/edgesec/src/recap.c:567:7: error: variable 'pc' is used uninitialized whenever 'if' condition is true [-Werror,-Wsometimes-uninitialized]
>   if ((eloop = eloop_init()) == NULL) {
>       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/runner/work/edgesec/edgesec/src/recap.c:608:14: note: uninitialized use occurs here
>   close_pcap(pc);
>              ^~
> /home/runner/work/edgesec/edgesec/src/recap.c:567:3: note: remove the 'if' if its condition is always false
>   if ((eloop = eloop_init()) == NULL) {
>   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /home/runner/work/edgesec/edgesec/src/recap.c:572:3: note: variable 'pc' is declared here
>   struct pcap_context *pc = NULL;
>   ^
> 1 error generated.
> ```

 
```suggestion
  struct pcap_context *pc = NULL;

  if ((eloop = eloop_init()) == NULL) {
    log_error("eloop_init fail");
    goto process_pcap_capture_fail;
  }
```

aloisklink (approved):
Looks good!

src/recap.c
mereacre: It doesn't it make it faster in the direct sense. But it decreases the packets dropped by the libpcap. So, my first estimates is that if one saves directly into the DB libpcap drops around 30% of packets, which is not good. The committing reduces the packet drop considerable. I will have to measure it.

mereacre: You can run the code on your PC and see many packets it drops (https://github.com/nqminds/edgesec/blob/recap-pcap/src/recap.c#L495). Eventually I will have to do similar modifications to the header middleware.

src/recap.c
mereacre: I can change to bool. But I'm an old timer :) Should have used char instead.

src/recap.c
mereacre: ok. I will merge this one first and then #116.

src/capture/pcap_service.h
mereacre: Done

src/recap.c
mereacre: Done.

src/recap.c
mereacre: Done

PR #350, opened 2022-11-25
Corrects the segmentation fault on capture thread closure.

mereacre:
Corrects the segmentation fault on capture thread closure.

Added:
 - Modified the tests to run the capture thread
 - Modified the main edgesec test to run the capture thread.


tests/capture/test_capture_service.c
```diff
@@ -147,23 +149,20 @@ void capture_config(struct capture_conf *config) {
...
-  struct capture_middleware_context context;
-
-  strcpy(context.ifname, "wlan0");
-  capture_config(&context.config);
+static void test_run_capture_thread(void **state) {
+  (void)state;
 
-  struct eloop_data *eloop = os_zalloc(sizeof(struct eloop_data));
-  assert_non_null(eloop);
+  char *ifname = "wlan0";
```
aloisklink, line 155: Use the following if this string will be modified:

```suggestion
  char ifname[] = "wlan0";
```

Or use the following if this string will not be modified:

```suggestion
  const char *ifname = "wlan0";
```

aloisklink (approved):
Looks good!

Minor comment, it would be nice if the commit description that fixed the segmentation fault said that it fixed the segmentation fault.

It makes `git bisect` and other `git` commands easier, especially when trawling through the git log looking for the commit that broke/fixed something.

mereacre:
Thanks for the feedback. This commit https://github.com/nqminds/edgesec/pull/350/commits/d5d50dd06f2c72cad75af9290b26e803f6058371 fixes the segfault. Will add more description in the future.

PR #351, opened 2022-12-01
Update control

mereacre:
Change the minimal version of cmake to 3.19.0 to allow presets.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: net
 Priority: optional
 Maintainer: Alois Klink <alois@nquiringminds.com>
 Build-Depends: debhelper-compat (= 12),
-    git, ca-certificates, cmake (>=3.15.0),
+    git, ca-certificates, cmake (>=3.19.0),
     doxygen, graphviz, texinfo,
     libnl-genl-3-dev, libnl-route-3-dev,
     autoconf, flex, bison,


mereacre:
The issue that I have is that we are using mainly presets. I don't expect other developers not to use presets and version 3.16 is pretty much useless for us.

I know snap has a more recent version of cmake. Can we integrate with snap?

aloisklink:
> The issue that I have is that we are using mainly presets. I don't expect other developers not to use presets and version 3.16 is pretty much useless for us.

True, true, but on the other side, the Debian build preset is pretty different from the rest of the other CMake development presets, so it's not too much trouble manually specify the `-DMY_VAR=VALUE` for it (I believe we need to do the same thing for OpenWRT buildroot).

> I know snap has a more recent version of cmake. Can we integrate with snap?

`snap` is great for applications, but it's pretty awful to use for automated builds. The entire point of the Debian build is that it uses the Debian system to build, so that we can make our own https://help.launchpad.net/Packaging/PPA or even add our package to Debian/Ubuntu once it gets more stable.

Plus, I don't even know if `snap` supports `armhf` (e.g. old raspberry pis/Turris Omnia), since Ubuntu doesn't support them. However, the Debian project still supports them.

Plus, if we want to publish `edgesec` as a `snap`, we'd have to ask for Ubuntu's permission, there's no way to self-host a snap, as far as I'm aware.

mereacre:
@aloisklink What's the decision on this PR? Do we close it and later on we add the snap integration?

aloisklink:
> Do we close it and later on we add the snap integration?

That's my gut feeling. But I'll be honest, I still think we should keep deb integration too, for Debian/light-weight Ubuntu distros.

It would also be nice to set up an edgesec PPA too so that we can have a super easy way to install edgesec on Debian/Ubuntu (and do automatic updates).

PR #352, opened 2022-12-02
fix: check if we are piping and not saving to db

mereacre:


src/recap.c
```diff
@@ -551,15 +551,19 @@ void eloop_tout_header_handler(void *eloop_ctx, void *user_ctx) {
   struct recap_context *pctx = (struct recap_context *)user_ctx;
 
   if (is_packet_queue_empty(pctx->pq) < 1) {
-    log_trace("Commiting packets to %s database", pctx->out_path);
-    if (execute_sqlite_query(pctx->db, "BEGIN IMMEDIATE TRANSACTION") < 0) {
-      log_error("Failed to capture a lock on db %s, ignoring.", pctx->out_path);
+    if (!pctx->pipe) {
+      log_trace("Commiting packets to %s database", pctx->out_path);
+      if (execute_sqlite_query(pctx->db, "BEGIN IMMEDIATE TRANSACTION") < 0) {
+        log_error("Failed to capture a lock on db %s, ignoring.", pctx->out_path);
```
aloisklink, line 557: Should we make this a `log_warn` if we can safely ignore this?

```suggestion
        log_warn("Failed to capture a lock on db %s, ignoring.", pctx->out_path);
```

src/recap.c
```diff
@@ -551,15 +551,19 @@ void eloop_tout_header_handler(void *eloop_ctx, void *user_ctx) {
...
+      }
     }
 
     save_packets_from_queue(pctx);
```
aloisklink, line 561: I'm a bit confused on to why we need this change.

As far as I'm aware, `save_packets_from_queue` is non-blocking, e.g. it reads as many packets as are available in the queue, then it returns once the queue is empty.

https://github.com/nqminds/edgesec/blob/20fd08788904519fa441375420891fc48c268737/src/recap.c#L536-L548

This means that either:

1. the database will be occasionally unlocked when the queue becomes empty, or
2. the queue fills up faster than the queue can be emptied, and removing these TRANSACTION statements will make the issue even worse.
  (maybe we can instead do something like `save_packets_from_queue(pctx, max_packets=128);` to force the function to eventually return in that case).

mereacre: There's the option to pipe as protobuf messages or to save into the sqlite db not as protobuf messages. The current fix fixes the issue when recap is trying to pipe no protobuf messages into the pipe. Therefore, you get the `if (!ctx->pipe)` check.

Notice that there are two options:
 - Pipe as protobofu messages into a named pipe
 - Save non protobuf messages (sqlite prepared statements) into the sqlite db

aloisklink: Aaah, I misunderstood :) My fault! I assumed you meant piped from libpcap, not piped to a named pipe!

src/recap.c
```diff
@@ -551,15 +551,19 @@ void eloop_tout_header_handler(void *eloop_ctx, void *user_ctx) {
   struct recap_context *pctx = (struct recap_context *)user_ctx;
 
   if (is_packet_queue_empty(pctx->pq) < 1) {
-    log_trace("Commiting packets to %s database", pctx->out_path);
-    if (execute_sqlite_query(pctx->db, "BEGIN IMMEDIATE TRANSACTION") < 0) {
-      log_error("Failed to capture a lock on db %s, ignoring.", pctx->out_path);
+    if (!pctx->pipe) {
+      log_trace("Commiting packets to %s database", pctx->out_path);
+      if (execute_sqlite_query(pctx->db, "BEGIN IMMEDIATE TRANSACTION") < 0) {
+        log_warn("Failed to capture a lock on db %s, ignoring.", pctx->out_path);
```
aloisklink, line 557: One last change to make `clang-format` happy :)

```suggestion
        log_warn("Failed to capture a lock on db %s, ignoring.",
                 pctx->out_path);
```

aloisklink (approved):
Looks good to me!

PR #356, opened 2022-12-05
Add initial support for FreeBSD

aloisklink:
Add edgesec support for FreeBSD.

Currently, edgesec can be compiled for FreeBSD, however, the following tests fail:

```
93% tests passed, 3 tests failed out of 46
Total Test time (real) =   2.79 sec
The following tests FAILED:
	  2 - test_edgesec (Failed)
	 40 - test_dnsmasq (Failed)
	 46 - test_mdns_service (Failed)
```

Additionally, compiling edgesec only works with `USE_GENERIC_IP_SERVICE`, which calls the `ip` command at runtime. However, FreeBSD does not have an `ip` command, so even though this network service compiles, it is impossible for it to run as it is right now.

[Cirrus CI](https://cirrus-ci.com/github/nqminds/edgesec) is currently used for FreeBSD CI.
I've skipped the above tests to that CI passes :)


mereacre (approved):
Looks grear.

mereacre (approved):

PR #360, opened 2022-12-09
Fix `-Wformat-security` errors

aloisklink:
Debian builds compile with `-Wformat-security` enabled.

This PR enables this flag, and fixes those errors.

See <https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html> for docs.

This caught a potential truncation bug in `dnsmasq.c` and a potential buffer overflow in `sqlhook.c`


mereacre (approved):
Very nice.

PR #364, opened 2022-12-09
build(cmake): use add_compile_options not C_FLAGS

aloisklink:
CMake does not recommend modifying `CMAKE_C_FLAGS` directly if we can help it. Instead, we should be using `add_compile_options()`.

This also means that we no longer overwrite `CMAKE_C_FLAGS`, which will mean that `edgesec` works better with other build systems (e.g. pdebuild).

Depends on #360 and #363 (since we no longer overwrite `CMAKE_C_FLAGS`, some of the debian build's C_FLAGS cause errors that need to be fixed).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,11 @@ else()
   # only use standard C and opt-in to non-standard includes
   set(CMAKE_C_EXTENSIONS OFF)
 endif()
-set(CMAKE_C_FLAGS "-Wunused-variable -Wall -Wextra")
+add_compile_options(
+  $<$<COMPILE_LANGUAGE:C>:-Wunused-variable>
+  $<$<COMPILE_LANGUAGE:C>:-Wall>
+  $<$<COMPILE_LANGUAGE:C>:-Wextra>
+)
 
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)


mereacre (commented):
From cmake docs:
The preprocessor definitions are added to the [COMPILE_DEFINITIONS](https://cmake.org/cmake/help/latest/prop_dir/COMPILE_DEFINITIONS.html#prop_dir:COMPILE_DEFINITIONS) directory property for the current CMakeLists file. They are also added to the [COMPILE_DEFINITIONS](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_DEFINITIONS.html#prop_tgt:COMPILE_DEFINITIONS) target property for each target in the current CMakeLists file.

In the cheri-pure branch I add compiler flags `CMAKE_C_FLAGS` in `cheri-cross.cmake` file located in CMakeToolchains. Now the question is will they be applied to the entire project?

mereacre (approved):

aloisklink:
> In the cheri-pure branch I add compiler flags `CMAKE_C_FLAGS` in `cheri-cross.cmake` file located in CMakeToolchains. Now the question is will they be applied to the entire project?

Yes, CMake Toolchain files are applied globally. The `add_compile_options()` will only apply to targets declared after the `add_compile_options()` call, but this PR adds those changes before everything is declared, so that's fine.

PR #368, opened 2022-12-12
Fix cross-compiling libs using autotools with cheribuild

aloisklink:
Improves the flags passed to autotools (used by libuuid, libmnl, libsqlite) to get edgesec compiled using [cheribuild](https://github.com/CTSRD-CHERI/cheribuild).

To do this, we need to pass `CFLAGS` (C compiler flags) and `LDFLAGS` (linker flags).

Additionally, `util-linux` uses [`SOLIB_LDFLAGS`](https://github.com/util-linux/util-linux/blob/5099e004b486e7c4f7153abc9843f49c1593fe66/configure.ac#L2721-L2722) to control linking for libraries.

However, these flags are first passed through [`libtool`](https://www.gnu.org/software/libtool). This causes issues when linking for ARM Morello PureCap on CheriBSD using cheribuild, as `libtool` clobbers the `-target aarch64-unknown-freebsd13` parameter.

Instead, we need to prepend the `-target aarch64-unknown-freebsd13` with `-XCClinker`, as that tells libtool to:

> Pass a link-specific flag to the compiler driver (CC) during linking

See https://www.gnu.org/software/libtool/manual/libtool.html#Link-mode

### DIFF
--- a/lib/mnl.cmake
+++ b/lib/mnl.cmake
@@ -28,7 +28,7 @@ elseif(BUILD_MNL_LIB)
       <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> "--host=${target_autoconf_triple}"
       # use position independent code, even for static lib, in case we want to make shared lib later
       --with-pic=on
-      "CC=${CMAKE_C_COMPILER}"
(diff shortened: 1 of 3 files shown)


mereacre (approved):
Great.

PR #369, opened 2022-12-12
build(test): allow cross-compiling tests

aloisklink:
Building tests is now controlled by the `BUILD_TESTING` flag, which is provided by `CTest`.

This allows us to cross-compile tests, which is useful for `cheribuild`, which has a QEMU emulator for running tests. This is also the standard way of doing things in CMake, see this guide to modern CMake: https://cliutils.gitlab.io/modern-cmake/chapters/testing.html


mereacre (approved):
Great.

PR #370, opened 2022-12-12
Compile edgesec `src/` dir with `-Wpedantic` and fix errors

aloisklink:
Compile with the `-Wpedandtic` flag in the edgesec `src/` directory.

`-Wpedandtic` checks for a lot more C errors that are often allowed by the GCC/Clang compilers, but are technically incorrect in ISO C.

However, it's worth enabling, as since we are compiling for CheriBSD, we want to follow standard C as much as possible.

We need to disable the `-Wpedandtic` flag when compiling `nl`, however. This is because `nl` includes `linux/netlink`, which includes some 0-length arrays, which are invalid in ISO C.

##### [fix(middlewares_list): fix invalid ISO C syntax](https://github.com/nqminds/edgesec/commit/dace5eee7fd76a43cc4a8ef828b989c580418249) 

In ISO C < C23, using an empty initializer list is illegal.
Instead, we can just make a pointer that points to NULL.

##### [refactor(dns_decoder): fix pointer types](https://github.com/nqminds/edgesec/commit/a14198ff724c40abaea07c690af5374bbfc6fda6) 

In ISO C, doing pointer arthemtic with a `(void *)` pointer is
undefined. If we instead switch to a `(char *)`, it's all fine!

I've also used ptrdiff_t and size_t when appropriate, instead of
basic `int` types.

##### [refactor(packet_decoder): fix pointer types](https://github.com/nqminds/edgesec/pull/370/commits/ec4893b6242c07edc58e9730ae2ab06c3dc2a470) 

In ISO C, doing pointer arthemtic with a `(void *)` pointer is
undefined.

##### [refactor(mdns_decoder): fix pointer types](https://github.com/nqminds/edgesec/pull/370/commits/0aa02878242141a7d02698bf9304502f2812dbbd) 

In ISO C, doing pointer arthemtic with a `(void *)` pointer is
undefined.

##### [refactor(radius): fix mac_conn_fn type](https://github.com/nqminds/edgesec/pull/370/commits/154b071ab3b34c766e8cce4b3420b407cd9f0897) 

In ISO C, casting from a function pointer to a void pointer is
undefined. Some platforms (e.g. z/OS or MS-DOS) have different sized
pointers for functions and data.

##### [refactor(ap_service): fix run_ap callback type](https://github.com/nqminds/edgesec/pull/370/commits/9e411ed3d799fa6986b5dc2ac1aa459453f90239)

In ISO C, casting from a function pointer to a void pointer is
undefined. Some platforms (e.g. z/OS or MS-DOS) have different sized
pointers for functions and data.

To get around this, we can instead pass run_ap() a pointer to a
struct that has a pointer to a function.

**This commit had the most changes, but they're still pretty minor**.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,6 +3,7 @@ add_compile_options(
   $<$<CONFIG:Debug>:-Werror>
   # Warn/error if using a non-const pointer to a string literal
   $<$<COMPILE_LANGUAGE:C>:-Wwrite-strings>
+  $<$<COMPILE_LANGUAGE:C>:-Wpedantic>
   $<$<COMPILE_LANGUAGE:C>:-Wformat-security> # required by debian builds
 )
 
--- a/src/ap/ap_service.c
+++ b/src/ap/ap_service.c
@@ -31,10 +31,6 @@
 #define AP_STA_DISCONNECTED "AP-STA-DISCONNECTED"
 #define AP_STA_CONNECTED "AP-STA-CONNECTED"
 
-typedef void (*ap_service_fn)(struct supervisor_context *context,
-                              uint8_t mac_addr[],
-                              enum AP_CONNECTION_STATUS status);
-
 int ping_ap_command(struct apconf *hconf) {
   char *reply = NULL;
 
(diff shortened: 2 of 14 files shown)


mereacre (approved):
The code is more cleaner now.

PR #372, opened 2022-12-12
test: fix casting ints to pointers

aloisklink:
For some our tests, in order to create some fake values, we were casting integers to pointers, which Cheri LLVM does not like.

Instead, we can use `mock_ptr_type()` to return a pointer type directly.

See: https://api.cmocka.org/group__cmocka__mock.html#gac267dacf70a0f82aa1521b01a1c98238


mereacre (approved):
Great.

PR #373, opened 2022-12-13
Optimize the `os_zalloc` and `os_calloc` functions by using `calloc`

aloisklink:
Instead of using our own implementation of calloc, we should use the C standard library version of [`calloc`](https://en.cppreference.com/w/c/memory/calloc).

This is because it will likely be faster, as instead of manually zero-ing memory in edgesec, the OS can just return pre-zeroed memory for us. See https://blogs.fau.de/hager/archives/825 for an example.

Additionally, by default, Linux uses overcommits memory using copy-on-write for memory allocation.

E.g., if you do `calloc(1, 2**40); // 1 GiB)`, Linux/MMU will allocate all 1 GiB of the memory to a single 4 KiB zero-ed page. Linux will then create a new memory page whenever you modify a page (so by default, `calloc(1, 2**40);` will use up 1 GiB of virtual RAM, and no physical RAM).

Physical RAM will only be used for any pages that you actually modify.

### DIFF
--- a/src/utils/allocs.c
+++ b/src/utils/allocs.c
@@ -18,12 +18,7 @@
 
 #include "allocs.h"
 
-void *os_zalloc(size_t size) {
-  void *n = os_malloc(size);
-  if (n != NULL)
-    os_memset(n, 0, size);
-  return n;
-}
+void *os_zalloc(size_t size) { return os_calloc(size, 1); }
 
 void *os_memdup(const void *src, size_t len) {
   void *r = os_malloc(len);
--- a/src/utils/allocs.h
+++ b/src/utils/allocs.h
@@ -29,25 +29,6 @@
  */
 void *os_zalloc(size_t size);
 
-/**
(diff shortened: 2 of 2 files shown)


mereacre (approved):
Interesting. Learned something today. I didn't use calloc at all in the code. But yes we can definitely use it of zalloc.

It will be useful to check the performance of the capture service with the new zalloc.

aloisklink:
> Interesting. Learned something today. I didn't use `calloc` at all in the code. But yes we can definitely use it of `zalloc`.

Basically, we should never use `zalloc`, since it's non-standard.

On some platforms, `calloc()` also prevents overflows. E.g., if you do `zalloc(big_number * big_number)`, you might get some of that magic undefined behavior if `big_number * big_number` overflows. `calloc(big_number, big_number)` sometimes checks this for you (although, this check might slow down `calloc()` vs `zalloc()`, so :shrug:)

PR #380, opened 2023-01-09
Radius erp

mereacre:
This PR incoroporates the ERP server into Radius and adds the vlan mechanism.

Changes made:
 - Loads of refactoring
 - todo


aloisklink:
I'm currently trying to `rebase` this branch onto `main` **without** the changes from #376.

**Current progress:** (2023-02-03)

This is going to be a bit of struggle to `git rebase`, especially since there's `96` commits to go through (+ some commits delete, then re-add files, which `git` isn't good with).

The command that I'm using for testing currently is:

```bash
git rebase --onto main 1da297dd3157ca8670a7a5e06174ebba7075d4d7 --exec 'rm -rf build/*/_deps/eloop_download-* && ( for PRESET in "linux" "linux-with-crypt" "linux-with-protobuf" "linux-with-example-middlewares"; do cmake --preset "$PRESET" && nice -n19 cmake --build --preset "$PRESET" -j=$(nproc) && { if ctest --list-presets | grep "\"$PRESET\""; then ctest --preset "$PRESET" --output-on-failure; fi } && cmake --install "./build/$PRESET" --prefix "./tmp/$PRESET" || exit 1; done ) && echo "all presets passed"'
```

mereacre:
It will not compile on all presets unless we solve the eloop naming issue.

PR #386, opened 2023-01-10
perf(os): improve performance of os_strlcpy (300% speed increase!)

aloisklink:
Implement `os_strlcpy` in terms of `memccpy()` and `strlen()`.

For most C libraries, `memccpy()` and `strlen()` are well optimized,
e.g. glibc uses AVX2 SIMD instructions, see https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/multiarch/strlen-avx2.S;h=b9b58ef599c19d2b13f9ff27c8af4bfbffecb735;hb=f049f52dfeed8129c11ab1641a815705d09ff7e8

I did a quick test doing `os_strlcpy(buffer, "my test "string");` 1'000'000'000 times and the old implementation took ~20 seconds, the new implementation takes ~5 seconds (4x speedup).

This change also fixes the `-Wconversion` warning thrown by both GCC and Clang.

### DIFF
--- a/src/utils/os.c
+++ b/src/utils/os.c
@@ -191,27 +191,18 @@ int hexstr2bin(const char *hex, uint8_t *buf, size_t len) {
   return 0;
 }
 
-size_t os_strlcpy(char *dest, const char *src, size_t siz) {
-  const char *s = src;
-  size_t left = siz;
+size_t os_strlcpy(char *restrict dest, const char *restrict src, size_t siz) {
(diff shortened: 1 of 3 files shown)


mereacre (approved):
Impressive speedup.

aloisklink:
> Impressive speedup.

It's because the old implementation was using standard C. `glibc` only works with GCC, and uses SIMD instructions, so they can break the C standard.

PR #393, opened 2023-01-27
refactor(eloop): add `edgesec_` prefix to funcs

aloisklink:
Add the `edgesec_` namespace prefix to all eloop functions.

We're having difficulties with using the libeap library, because it also uses eloop, but a different version of eloop. This is causing linking errors.

Unfortunately, we're not using C++, so we can't just wrap everything in a namespace, so we have to manually namespace the functions by adding our own custom prefix.

---

Before merging this PR, do you want to test this locally with `libeap` to see if it works?

By the way, @mereacre, if you prefer to see the `.patch` in GitHub's viewer, see https://github.com/nqminds/hostap/commit/38b00e51cb253b0955a10ff3a0e0cc9d12b8409c (I've made a private fork of `hostap` in our @nqminds organization)

To be honest, instead of dealing with `.patch` files in this repo, we could just download from own forked version of `hostap` at https://github.com/nqminds/hostap/commit/38b00e51cb253b0955a10ff3a0e0cc9d12b8409c, but :shrug:, I don't know if it's worth the effort in changing the system we have now (plus, then we'd have to make https://github.com/nqminds/hostap public).

### DIFF
--- a/lib/eloop.cmake
+++ b/lib/eloop.cmake
@@ -1,14 +1,19 @@
 if (NOT (BUILD_ONLY_DOCS))
     include(FetchContent)
 
-    # To generate or modify these patch files, do:
-    # cd "${eloop_download_SOURCE_DIR}"
-    # git init
-    # git add . && git commit -m "initial commit"
-    # git am ~/edgesec/lib/eloop/patches/*.patch
+    # To generate or modify these patch files, do the following to recreate
+    # the hostapd code in a git repo:
+    #
+    #     cd "${eloop_download_SOURCE_DIR}"
+    #     git init
+    #     git add . && git commit -m "initial commit"
+    #     git am ~/edgesec/lib/eloop/patches/*.patch
+    #
+    # To make this easier, you can use https://github.com/nqminds/hostap/tree/edgesec
+    # which may already has these patches applied.
     #
     # Then you can use `git rebase` to modify your git history.
-    # When done, do `git format-patch <FIRST_COMMID_ID>` to remake the patches
+    # When done, do `git format-patch hostap_2_10` to remake the patches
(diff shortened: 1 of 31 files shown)


mereacre:
I would prefer the following:
1. As you suggested we move the patches to a separate repo pof hostapd. We have already one setup on nqminds. The files there will be already patched.
2. Instead of `edgesec_eloop_*` we should use `evloop_*`. I don't want to use the name `edgesec` in the code as we might one day replace the name

aloisklink:
> Instead of `edgesec_eloop_*` we should use `evloop_*`. I don't want to use the name edgesec in the code as we might one day replace the name

`evloop` is already used by a bunch of projects (including tor).

If you don't want to use `edgesec` as a prefix, should we just create a persistent internal codename for `edgesec`? It's pretty standard to have a different internal codename for developers than the public name that marketing decides on (e.g. see [Microsoft](https://en.wikipedia.org/wiki/List_of_Microsoft_codenames) and [Apple](https://en.wikipedia.org/wiki/List_of_Apple_codenames))

If we're doing with a greek theme, we could try `moirae_` (after https://en.wikipedia.org/wiki/Moirai)

(or since modern C99 allows unicode in identifiers, we could use use `mœræ_` to make it almost impossible that anyone else would try to copy us.)

(or even `𓃣_eloop_` since nobody uses the Egyptian hieroglyphics Unicode characters (it can mean "he who is over the secrets" [𓃣](https://commons.wikimedia.org/wiki/Category:Recumbent_canine_(hieroglyph))) :laughing:)

In the long-run, we should probably switch to using something like https://github.com/libevent/libevent instead of our forked version.

mereacre:
I agree in the long run we should use the event loop library.

How about we replace `edgesec_eloop_*` with `edge_eloop__*`?

mereacre (approved):

PR #396, opened 2023-01-30
ci: remove deprecated `set-output` commands

aloisklink:
[GitHub Actions has deprecated the `set-output` command.][1]

They expect to disable the command fully in May 2023.

Migrating is pretty easy, all we need to do is replace the `echo "::set-output name={name}::{value}"` syntax with the new `echo "{name}={value}" >> $GITHUB_OUTPUT` syntax.

[1]: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### Notes to reviewers

@AshleySetter, don't worry, despite being in the `edgesec` repo, this isn't C code :wink:

I'm guessing you also program in GitHub Actions' YAML DSL (YAML is an amazing programming language), so you might be interested in this change, if you haven't heard about it already.

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -60,9 +60,9 @@ jobs:
         run: |
           if ctest --list-presets | grep -s "${{ matrix.cmake-preset }}"; then
             # ctest --preset "${{ matrix.cmake-preset }}" --output-on-failure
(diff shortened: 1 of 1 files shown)


AshleySetter (approved):
LGTM!

PR #398, opened 2023-01-30
chore(deps): update hashes of GitHub downloads

aloisklink:
GitHub is very nice, and has decided to recreate all of the .tar.gz files with slightly worse compression, so that we can have the fun of updating all of our hashes! Isn't that nice :)

(I've confirmed that running `gunzip .tar.gz` outputs the same `.tar` file, so it seems like it's just slightly worse compression).

According to GitHub staff, all of the `/archive/refs/tags/$tag` URLS should be stable, but I guess they were stable until today :shrug:.

See https://twitter.com/thesayynn/status/1620129657977987073

### DIFF
--- a/.github/workflows/create-debs.yml
+++ b/.github/workflows/create-debs.yml
@@ -60,9 +60,9 @@ jobs:
         run: |
           if ctest --list-presets | grep -s "${{ matrix.cmake-preset }}"; then
             # ctest --preset "${{ matrix.cmake-preset }}" --output-on-failure
-            echo "::set-output name=tested::true"
+            echo "tested=true" >> $GITHUB_OUTPUT
           else
-            echo "::set-output name=tested::false"
+            echo "tested=false" >> $GITHUB_OUTPUT
           fi
       - name: Code Coverage
         # code coverage only works if ctest works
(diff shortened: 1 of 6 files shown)


aloisklink:
@AshleySetter, any chance you want to do a review? I'm guessing this might be relevant to other projects too, (e.g. `npm install` from a git URL like https://github.com/nqminds/many-secured/blob/d9d4161230d530665e125be557e694091e040f11/package-lock.json#L5555)?

mereacre (approved):

mereacre (approved):

aloisklink:
Hi @mereacre, looks like GitHub changed course and undid their changes, see https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/

(side-note, it looks like this was due to a recent change in `git`'s compression algorithm, see https://github.com/git/git/commit/4f4be00d302bc52d0d9d5a3d4738bb525066c710)

They said "GitHub doesn’t guarantee the stability of checksums for automatically generated archives.", but Hyrum's Law means if you expose an API, it becomes part of the API.

image

_Obligatory XKCD: https://xkcd.com/1172/_

PR #413, opened 2023-02-17
refactor: inline os_zalloc for internal linkage

aloisklink:
Change `os_zalloc` to use `static inline` so that it uses internal linkage.

The libeap library also has it's own `os_zalloc`, which is causing linking errors. We can easily avoid this by inlining the edgesec version of `os_zalloc`. The function is literally just one-line of code, so inlining it should basically change nothing.

---

I made this commit to replace https://github.com/nqminds/edgesec/pull/380/commits/93d8c8a827c243cde9ffd4df20e5d6f1a77a02aa, since there's no need to rename `os_zalloc` to `sys_zalloc` if we can just inline the function.

### DIFF
--- a/src/utils/allocs.c
+++ b/src/utils/allocs.c
@@ -18,8 +18,6 @@
 
 #include "allocs.h"
 
-void *os_zalloc(size_t size) { return os_calloc(size, 1); }
-
 void *os_memdup(const void *src, size_t len) {
   void *r = os_malloc(len);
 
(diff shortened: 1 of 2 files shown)


mereacre (approved):
Eventually we will have to replace again sys_"name" to os_"name".

aloisklink:
> Eventually we will have to replace again sys_"name" to os_"name"

I've tried to always add `#define os_"name"(...args) edge_os_"name"(...args)`, so that we only need to change the `os.h` and `os.c` file, and we don't need to change where the functions are called. It makes the `git diff` a lot smaller, and means that there are much much less merge conflicts.

PR #432, opened 2023-02-20
Use `__maybe_unused` compiler attribute to fix flakey code-coverage in `test_edgesec`

aloisklink:
For some reason, these lines are sometimes not run by tests, and because of that, we get a code-coverage warning:

image

https://github.com/nqminds/edgesec/blob/9b53dea608c87cb26cc5c25091f03f13afd8dfdc/tests/test_edgesec.c#L44-L45

I would have thought that these no-op statements are optimized out, even when compiling in debug mode, but I guess sometimes they are kept in?

To fix this, I've taken the `__maybe_unused` macro from PR #380 and stuck it in a new file called `src/utils/attributes.h`, which we can use for compiler attributes.
([C23 has built-in support for attributes](https://en.cppreference.com/w/c/language/attributes) like [`[[maybe_unused]]`](https://en.cppreference.com/w/c/language/attributes/maybe_unused), but since we're not using that, we're stuck with doing complicated compiler-specific checks in massive header files)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -67,6 +67,6 @@ set_target_properties(edgesec PROPERTIES
 
 if (USE_CAPTURE_SERVICE)
   add_executable(edgesec-recap edgesec-recap.c)
-  target_link_libraries(edgesec-recap PRIVATE capture_service protobuf_middleware packet_queue packet_decoder sqlite_header os log SQLite::SQLite3 eloop::eloop)
+  target_link_libraries(edgesec-recap PRIVATE capture_service protobuf_middleware packet_queue packet_decoder sqlite_header attributes os log SQLite::SQLite3 eloop::eloop)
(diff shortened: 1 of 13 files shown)


mereacre (approved):
Great.

We should also add https://github.com/nqminds/edgesec/blob/main/src/utils/os.h#L36 i.e.,
```c
__attribute__((packed))
```
to the `attributes.h` file.
```

aloisklink:
> Great.
> 
> We should also add https://github.com/nqminds/edgesec/blob/main/src/utils/os.h#L36 i.e.,
> 
> ```c
> __attribute__((packed))
> ```
> 
> to the `attributes.h` file.

Good idea, I've done it in https://github.com/nqminds/edgesec/pull/432/commits/f770f44fc08594d57f6feb36b5fdfe09afb69049

PR #433, opened 2023-02-20
test(eloop): test if 1mil µs overflows to 1s

aloisklink:
Test whether 1 million microseconds overflows to 1 second when running `edge_eloop_register_timeout()`.

This was causing code-coverage issues, because this path was occasionally not run in the current tests.

image

(source-code available at https://w1.fi/cgit/hostap/tree/src/utils/eloop.c?h=hostap_2_10#n791)


mereacre (approved):
I like this one CↈↃµs :)

PR #440, opened 2023-02-21
feat(common): add wpa_hexdump_ascii function

aloisklink:
Add `wpa_hexdump_ascii` to match the API of `wpa_hexdump_ascii()` from hostap.

Please note that this function acts differently from the hostap version of the function:
  - It only prints the first 16 bytes.
  - It prints all bytes as hex, never as ASCII.

---

Adapted from commit https://github.com/nqminds/edgesec/commit/dff384c0c21e41af7da63a45bfb3574322a6f9c8.

My changes compared to https://github.com/nqminds/edgesec/commit/dff384c0c21e41af7da63a45bfb3574322a6f9c8:
  - Added documentation
  - Use `__maybe_unused` (from https://github.com/nqminds/edgesec/commit/5a3a8bf22e4976c8fe9ff37daa1e221fd299da01) to mark the unused `level` param.
  - I've changed the number of bytes in the buffer to 33 instead of 32. This is because we need a byte for the `NUL`-terminator too (see https://github.com/nqminds/edgesec/pull/439/commits/24fe4cbd56740bc828bc2016652dad59b6b67c24 for more information), otherwise `printf_hex` causes a mid-byte truncation issue.

### DIFF
--- a/src/radius/common.h
+++ b/src/radius/common.h
@@ -155,6 +155,27 @@ static inline u32 WPA_GET_BE24(const u8 *a) {
 #define wpa_snprintf_hex(buf, buf_size, data, len)                             \
   printf_hex(buf, buf_size, data, len, 0)
 
+/**
+ * Prints the first 15-bytes of the given buffer with the given title.
+ *
+ * @param level priority level of the message
+ * @param title of for the message
+ * @param data buffer to be dumped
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Great.

PR #449, opened 2023-02-21
refactor(supervisor_config): include `radius_config`

aloisklink:
> **Warning** This PR is based on the follow PRs. Please merge those PRs first before reviewing this PR.
>   - #447

Add an `#include "../radius/radius_config.h"` to `supervisor_config.h`.

This file uses the `struct radius_conf` type since https://github.com/nqminds/edgesec/commit/24c49d59fa36b3fc8e97a24a0b0a199fdae18bf1, which is defined in `../radius/radius_config.h`, so we should include that file directly.

---

This was changed in https://github.com/nqminds/edgesec/commit/2704b75351c6f688450766efd33a2ccd67fa1ff1#diff-c92e4b6dccbcdfb60672a8b00773227eddcc9c4ba6ee87a3ac46dcef04b41b4cR19, but I'm not 100% sure why.

I think @mereacre did it because there's `struct radius_conf` in this file, so that's what I've stuck in this PR/commit description (if I'm wrong, let me know @mereacre).

### DIFF
--- a/src/supervisor/CMakeLists.txt
+++ b/src/supervisor/CMakeLists.txt
@@ -63,7 +63,7 @@ endif()
 
 add_library(supervisor_config INTERFACE)
 set_target_properties(supervisor_config PROPERTIES PUBLIC_HEADER "supervisor_config.h")
-target_link_libraries(supervisor_config INTERFACE SQLite::SQLite3 iface ap_config dhcp_config)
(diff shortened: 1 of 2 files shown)


mereacre:
I think I had some issue with recursive definitions. But as long as it compiles fine, shouldn't be no problem.

mereacre (approved):

Issue #435, opened 2023-02-21
`tests/test_edgesec.c` code-coverage is inconsistent


aloisklink:
**Describe the bug**

`tests/test_edgesec.c` code-coverage is inconsistent.

As an example, see https://github.com/nqminds/edgesec/pull/434#issuecomment-1438338979.

This PR doesn't change anything in `tests/test_edgesec.c`, but somehow `tests/test_edgesec.c` had a code-coverage change. This is consistently happening with our PRs, which is a bit annoying, since we sometimes get a false-positive ❌ saying that the code-coverage got worse, but it turns out it's just `tests/test_edgesec.c`

**Expected behavior**

`tests/test_edgesec.c` code-coverage is consistent.

**Screenshots**

![image](https://user-images.githubusercontent.com/19716675/220341245-7436565d-f3dd-4f3c-b238-c5bf5ed81566.png)

**Additional context**

It looks like the issue is this line: https://github.com/nqminds/edgesec/blob/595fe952283981e5fcec488d90e26fb0dfdc5c6e/tests/test_edgesec.c#L60

I'm guessing maybe some of the CI compilers might be using a different version of `gcc` that uses different optimisation? I have a WIP commit that will test this branch, which should fix this issue, see https://github.com/nqminds/edgesec/commit/80e1ee793da3eafe8bc8cc9121a685aaadfcfda9

aloisklink:
It looks like this is still an issue, as of https://github.com/nqminds/edgesec/commit/dd4f7a655a9c889ceb843e1be13e300c4b7e8a09.

The issue seems to be that when GCC outputs the tracefile showing the code-coverage, it ignores this line, because it has no assembly associated with it:

https://github.com/nqminds/edgesec/blob/39baecc8f37eee6f76bfe4ac8ca52d9f77bc4a32/tests/test_edgesec.c#L79

However, when Clang outputs their tracefile for code-coverage, it doesn't ignore this line. 

The following line of code (the branch above the `} else {` line) is tested but only on the OpenWRT test preset, which only uses GCC.

https://github.com/nqminds/edgesec/blob/39baecc8f37eee6f76bfe4ac8ca52d9f77bc4a32/tests/test_edgesec.c#L78

I think if we test this branch in Clang too, this should fix this issue.

PR #453, opened 2023-02-24
Improves `test_edgesec.c`

aloisklink:
Improves `test_edgesec.c` by making the following changes:

- [**test(edgesec): use `threads.h` in test_edgesec**](https://github.com/nqminds/edgesec/commit/a8d1aa93e6fbd86a662c67cc432a857f8ea7f27c) 

  Use `#include <threads.h>` (aka ISO C11 standard threads) instead of `#include <pthreads.h>` in `test_edgesec.c`
  Using C11 standard threads caught a double-free error, so it's already
better!
- [**test(edgesec): stop eloop threads gracefully**](https://github.com/nqminds/edgesec/commit/d76a2787da075bbc33d1177fe9336bf99ddbb82b)
  Currently, the eloop threads are never actually stopped. We call `edge_eloop_terminate()`, but this is not thread-safe.

  To fix this, we can use a threadsafe atomic from `<stdatomic.h>`, which is constantly checked by eloop, which then automatically closes the eloop when the atomic is set.

  This also allows us to call `thrd_join()` on the threads and check their return code.

  Basically, before this PR, calling `edge_eloop_terminate()` from the main test thread did nothing.

- [**test(edgesec): store error msg from ap thread**](https://github.com/nqminds/edgesec/commit/199b8cfc7c37fc3edaec24516c943b338916d902) 

  Currently, in `test_edgesec`, if the AP thread has an error, `fail_msg()` is called, which normally fails the CMocka test with the given error message. However, [CMocka **is not thread safe**][1].

  Instead, we can make the AP Thread store the error message in a buffer, and call `thrd_exit(EXIT_FAILURE)` to exit the thread with an error exit code.

  We can then check for this thread exitcode later on, when we call `thrd_join()`, e.g. in `teardown_edgesec_test()`.

- [**test(edgesec): test ap server thread failures**](https://github.com/nqminds/edgesec/commit/8f63f7abb87daa92e1942972b76449a1a4828958) 

  Add a test that tests whether the AP server thread fails correctly.

  Previously, our test code wasn't correctly failing on errors.

  This should also help fix flakey code-coverage results in `test_edgesec.c`, see https://github.com/nqminds/edgesec/issues/435.

  This test is pretty slow, as we need to wait 10 seconds for `writeread_domain_data_str()` to timeout. In the future, we should find a way to speed up this test (maybe by using a different function to send data to the AP server socket).

[1]: https://api.cmocka.org/

---

~@mereacre, if you've got any idea how we can easily replace this line, without having to wait 10 seconds for a timeout, let me know! https://github.com/nqminds/edgesec/blob/43f5b5284a81f7c0d344200a9adf86bb3ef56a56/tests/test_edgesec.c#L376-L381~ _Edit: Fixed in https://github.com/nqminds/edgesec/pull/453/commits/d6c0a99924b84f8f58c98972045d6828264da64b_

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -50,7 +50,7 @@ if (USE_RADIUS_SERVICE)
     LINK_LIBRARIES radius radius_client attributes sockctl runctl cmocka::cmocka log config Threads::C11Threads
   )
   target_compile_definitions(test_edgesec PRIVATE TEST_CONFIG_INI_PATH="${CMAKE_BINARY_DIR}/test-config.ini")
-  set_tests_properties(test_edgesec PROPERTIES TIMEOUT 10)
+  set_tests_properties(test_edgesec PROPERTIES TIMEOUT 15)
 endif()
 
 add_cmocka_test(test_runctl
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -50,7 +50,7 @@ if (USE_RADIUS_SERVICE)
     LINK_LIBRARIES radius radius_client attributes sockctl runctl cmocka::cmocka log config Threads::C11Threads
   )
   target_compile_definitions(test_edgesec PRIVATE TEST_CONFIG_INI_PATH="${CMAKE_BINARY_DIR}/test-config.ini")
-  set_tests_properties(test_edgesec PROPERTIES TIMEOUT 15)
+  set_tests_properties(test_edgesec PROPERTIES TIMEOUT 20)
 endif()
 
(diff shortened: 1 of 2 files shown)


mereacre (approved):
Great work. Indeed we will have to replace the timeouts eventually.

aloisklink:
> Great work. Indeed we will have to replace the timeouts eventually.

Yah, even 1 second is quite a lot when I'm running the tests 1000s of times when `git rebase`-ing :)

Plus, think about all of the CO2 we'd save by using less of Microsoft's/GitHub's server time :rofl: !!

Issue #425, opened 2023-02-17
Running `ctest` in parallel may cause `test_dnsmasq` to fail due to SIGHUP


aloisklink:
**Describe the bug**

Running tests with `ctest` only works when using a single process.

Running `ctest` with multiple threads (e.g. `ctest -j16`), has a high chance of causing `test_dnsmasq` to be killed via `SIGHUP`.

**To Reproduce**

1. First compile tests

    ```bash
    git checkout ba237fdcc13c9313be9aed6fdff60eba996115cf # current main
    cmake --preset "linux"
    cmake --build --preset "linux" -j=16
    ```

2. Run `ctest` with multiple threads:

    ```console
    me@me:~/edgesec$ ctest --preset "linux" --output-on-failure -j=16 # often fails!
    ...
    15/45 Test #36: test_hostapd .....................   Passed    0.01 sec
          Start 44: test_reflection_list
    16/45 Test #39: test_dnsmasq .....................SIGHUP***Exception:   0.02 sec
    [==========] tests: Running 7 test(s).
    [ RUN      ] test_define_dhcp_interface_name
    2023-02-17 18:25:53.745  ERROR dnsmasq.c:72: dconf param is NULL
    2023-02-17 18:25:53.745  ERROR dnsmasq.c:77: ifname param is NULL
    2023-02-17 18:25:53.745  ERROR dnsmasq.c:111: define_dhcp_interface_name: snprintf error.
    [       OK ] test_define_dhcp_interface_name
    [ RUN      ] test_generate_dnsmasq_conf
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:171: Number of substrings=5
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:178: vlanid=0
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:188: ip_addr_low=10.0.0.2
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:195: ip_addr_upp=10.0.0.254
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:202: subnet_mask=255.255.255.0
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:209: lease_time=24h
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:171: Number of substrings=5
    2023-02-17 18:25:53.745  TRACE test_dnsmasq.c:178: vlanid=1
[...]

PR #455, opened 2023-02-27
Makes `is_proc_app()` stricter, so that it only looks at the process name

aloisklink:
#### 🐛 The problem

`is_proc_app()` currently returns the `pid_t` if the proc_name is the PID's `/proc/[pid]/exe`'s basename, OR if the proc_name is anywhere in the `/proc/[pid]/cmdline`.

This is far far far too broad.

For example, if I run the following:

```bash
/usr/bin/bash -c 'echo "Modifying /etc/hostapd.conf"; sed ...'
```

`is_proc_app(pid_dir, "hostapd")` will return that PID, because `hostapd` is in the argv2 string!

This also causes a bunch of false positives. For example, `signal_process("dnsmasq", SIGHUP);` kills `test_dnsmasq`, and so prevents tests from being run in parallel (fixes https://github.com/nqminds/edgesec/issues/425).

And `is_proc_running("hostapd")` returns `true` for `test_hostapd`.

#### 👷 The fix

Instead, it makes a lot more sense only to check the basename of the argv0 string.

For example, when calling the following command, **ONLY** the following `proc_name`s will return a PID:
- "gcc"
- "x86_64-linux-gnu-gcc-11" (assuming that `/usr/bin/gcc` is a symlink
  to `/usr/bin/x86_64-linux-gnu-gcc-11`).

```bash
/usr/bin/gcc -option1 -option2 'arg that doesn't matter'
```

### 📏 Implementation details

I've done quite a bit of documentation/refactoring when trying to get these functions working properly, so please review this PR commit-by-commit!

### DIFF
--- a/src/utils/os.c
+++ b/src/utils/os.c
@@ -860,22 +860,32 @@ pid_t is_proc_app(const char *path, const char *proc_name) {
                       parsed_pid_long // check that `unsigned long` value didn't
                                       // overflow when converting to pid_t
     ) {
-      log_error("is_proc_app: Failed to parse the pid in %s", path);
+      // don't log anything, since we expect this to happen for many files in
+      // /proc
       return 0;
     }
   }
 
-  char exe_path[MAX_OS_PATH_LEN];
-  char cmdline_path[MAX_OS_PATH_LEN];
+  {
+    // check if the basename of the realpath of the process matches proc_name
+    char exe_path[MAX_OS_PATH_LEN];
+    snprintf(exe_path, MAX_OS_PATH_LEN - 1, "%s/exe", path);
+    char resolved_path_buffer[PATH_MAX];
+    char *resolved_path = realpath(exe_path, resolved_path_buffer);
+    if (resolved_path == NULL) {
+      // don't log anything, since many `/proc/[pid]/exe` point nowhere, if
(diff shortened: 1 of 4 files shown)


mereacre (approved):
Good catch.

aloisklink:
Btw, @mereacre, before we merge, do you remember why you added the function is `is_string_in_cmdline_file()`, where you checked whether the given `char *proc_name` was in any of `/proc/[pid]/cmdline` https://github.com/nqminds/edgesec/blob/750e6d1ba6b418ace01f01f883009ae724f37243/src/utils/os.c#L806-L829 

(seems to be added in commit https://github.com/nqminds/edgesec/commit/adf2242c6764b80da0b1fdb02bd0e2ce58d5cddf)

I just want to check that I'm not breaking anything by changing the function to only looking at the `basename(argv0)` (i.e. you weren't try  OpenWRT doesn't do something weird like `/usr/bin/python3 some-fake-dnsmasq-script-that-you-wantyedt-signal-because-it-has-dnsmasq-in-the-name.py`)

mereacre:
It shouldn't break anything. Its sole purpose is to kill running edgesec,
dnsmasq and hostapd processes.

PR #467, opened 2023-02-28
Add a new GCC attribute called `__must_free` to warn if pointers aren't deallocated

aloisklink:
Add the `__must_free` attribute, that in GCC 11 or later, is defined as:

`#define __must_free __attribute__((malloc(free, 1)))`.

Basically, it tells GCC that the result of these functions must:
 1. be checked for `NULL`
 2. must be deallocated by `free()`

Otherwise, when compiling code that uses these functions in GCC with the `-fanalyzer` option, you'll get a warning that looks a bit like this:

```
<source>:17:1: warning: leak of 'x' [CWE-401] [-Wanalyzer-malloc-leak]
  'my_func': events 1-2
    |
    |   15 |   char * x = my_malloc(sizeof(int));
    |      |              ^~~~~~~~~~~~~~~~~~~~~~
    |      |              |
    |      |              (1) allocated here
    |   16 |   printf("%p", x);
    |   17 | }
    |      | ~             
    |      | |
    |      | (2) 'x' leaks here; was allocated at (1)
    |
Compiler returned: 0
```

Ideally, we should add this attribute to every function in edgesec that returns a value that must be `free()`-ed.

### DIFF
--- a/src/utils/allocs.h
+++ b/src/utils/allocs.h
@@ -29,7 +29,7 @@
  * @param size Number of bytes to allocate
  * @return void* Pointer to allocated and zeroed memory or %NULL on failure
  */
-__must_check static inline void *os_zalloc(size_t size) {
+__must_free static inline void *os_zalloc(size_t size) {
   return calloc(size, 1);
 }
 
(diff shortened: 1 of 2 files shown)


mereacre (approved):
Super. I added the same attribute to the BRSKI repo.

PR #470, opened 2023-02-28
refactor(common): match wpa log levels to edgesec

aloisklink:
Match the log levels used by source-code taken from hostap with the log levels used by edgesec, instead of always just using `LOGC_TRACE`.

This means that messages from the radius server will now be logged by default, and they'll be colour-coded for severity.

@mereacre, I don't know how verbose the `radius_server` is. If it's too much logging compared to the rest of edgesec, you're welcome to push down the values of `enum hostap_log_level`, e.g. so that `MSG_ERROR` only equals to an edgesec `LOGC_WARN`, for instance.


mereacre (approved):
Let's go with the levels in your PR.

PR #480, opened 2023-03-06
Make the `cmd` parameter in `get_command_function() `const`

aloisklink:
Make the `cmd` parameter of `get_command_function()` `const`.

We only read this string and never change it, so we might as well make it `const`.

---

Side-note, all the `strcmp(...)` are a bit slow, since it's $O(n)$ (i.e. it scales linearly with the number of command types.

We could switch to using a hashmap (like [uthash.h](https://troydhanson.github.io/uthash/userguide.html)), which would be $O(1)$ (i.e. constant time with every command type), but if we only have 25 message types, I've got no idea how big the speedup will be :shrug:

Still, it's something to consider the more message types we add.

### DIFF
--- a/src/supervisor/cmd_processor.c
+++ b/src/supervisor/cmd_processor.c
@@ -971,7 +971,7 @@ ssize_t process_sign_blob_cmd(int sock, struct client_address *client_addr,
 }
 #endif
 
-process_cmd_fn get_command_function(char *cmd) {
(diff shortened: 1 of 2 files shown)


mereacre:
It is not worth the complexity because `get_command_function` it is called only for managing the network control layer. We could also replace the commands names with two latter names to speedup if needed.

PR #485, opened 2023-03-06
Fix `uwrt_lookup_package()` in `uci_wrt.c`

### LINKED ISSUE #481: Memory leak in `uci_lookup_section_ref()` function
It looks like the `uci_lookup_section_ref()` function has a memory leak.

`ti` is allocated with `calloc()`, but it's never `free()`-ed in the function, and as far as I'm aware, the pointer to `ti` is never passed out of this function.

https://github.com/nqminds/edgesec/blob/a3bcd1bdc25771312de474b5549d8f4e5bcd80dd/src/utils/uci_wrt.c#L76-L83

aloisklink:
Fix the `uwrt_lookup_package()` function in `uci_wrt.c`.

`uwrt_lookup_package()` is currently never adding anything to it's `list` param, which means that the count for anonymous section references is always 0. (this is also the cause of the memory leak in #481).

For example, given the following UCI file:

```conf
config rule
  option hi 'hi'

config rule
  option hi 'hi'

config rule
  option hi 'hi2'
```

Output:

| # | Current | Expected |
| -: | -- | -- |
|  0 | `@rule[0]` | `@rule[0]` |
|  1 | `@rule[0]` | `@rule[1]` |
|  2 | `@rule[0]` | `@rule[2]` |

To fix this, I've instead changed `uci_lookup_section_ref()` to use a [uthash.h](https://troydhanson.github.io/uthash/userguide.html) hashmap as a counter, instead of a custom linked-list.

Fixes #481

### Other changes

For the following functions, I've also made their parameters `const` when possible, add documentation of what the functions did, and made them `static`/gave them internal linkage, since they're never used outside of `uci_wrt.c`.

  - `uwrt_get_option()`
  - `uwrt_lookup_option()`
  - `uwrt_lookup_section()`

I've made the return string of `uci_lookup_section_ref()` a `const char *` instead of a `const char *` because it should **NOT** be `free()`-ed, as instead `uci_lookup_section_ref()` `malloc()`'s it's `malloc_buffer` param and that should be `free()`-ed instead.

Finally, I've fixed a double-`free()` error in `malloc_buffer` (originally named `typestr`).

- 1st `free()`:
https://github.com/nqminds/edgesec/blob/3b38b374c146d61207b705df084938b826b7e526/src/utils/uci_wrt.c#L93
- 2nd `free()`:
https://github.com/nqminds/edgesec/blob/3b38b374c146d61207b705df084938b826b7e526/src/utils/uci_wrt.c#L231

### Notes to reviewer (weird behaviour of `uwrt_get_option()`)

The behaviour of `uwrt_get_option()` is a bit weird. It almost exactly matches the syntax from the `uci show system` command-line, except: 

1. It doesn't quote the output.
2. List options are concatenated together.

For example:

```conf
# in file /etc/config/my_config
config rule 'section_name'
  option string_opt 'my-val'
  list list_opt 'list-val-1'
  list list_opt 'list-val-2'
```

The output for `uci_wrt.c` will look like:

```
my_config.section_name=rule
my_config.section_name.string_opt=my-val
my_config.section_name.list_opt=list-val-1list-val-2
```

However, if we ran `uci show my_config` on an OpenWRT box, we would instead get something like:

```
my_config.section_name='rule'
my_config.section_name.string_opt='my-val'
my_config.section_name.list_opt='list-val-1' 'list-val-2'
```

I've documented this behaviour, but we may want to change our `uci_wrt.c` behaviour to match that of the `uci` CLI tool, since I feel like we might have escaping issues with our version.

### DIFF
--- a/src/utils/uci_wrt.c
+++ b/src/utils/uci_wrt.c
@@ -292,8 +292,8 @@ static int uwrt_lookup_option(const struct uci_option *o, const char *sref,
  * @retval  0 On success.
  * @retval -1 On failure. The array may be partially filled in an error.
  */
-int uwrt_lookup_section(const struct uci_section *s, const char *sref,
-                        UT_array *kv) {
+static int uwrt_lookup_section(const struct uci_section *s, const char *sref,
+                               UT_array *kv) {
   const struct uci_element *e = NULL;
   const char *cname = s->package->e.name;
   const char *sname = (sref != NULL ? sref : s->e.name);
--- a/src/utils/uci_wrt.c
+++ b/src/utils/uci_wrt.c
@@ -134,7 +134,8 @@ uci_lookup_section_ref(struct uci_section *s,
       char *p = os_realloc(*typestr, maxlen);
       if (p == NULL) {
         log_errno("os_realloc");
-        os_free(*typestr);
+        // don't free() *typestr, next call to uci_lookup_section_ref()
+        // may reuse this value
         return NULL;
       }
 
--- a/src/utils/uci_wrt.c
+++ b/src/utils/uci_wrt.c
@@ -152,7 +152,30 @@ static char *uci_lookup_section_ref(struct uci_section *s,
   return ret;
 }
 
-char *uwrt_get_option(const struct uci_option *o) {
+/**
+ * @brief Gets the value of the given option.
+ *
+ * For list options, the value would be all the list entries concatenated
+ * together.
+ *
+ * For example, given the following UCI file:
+ *
+ * ```uci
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Great work.

aloisklink:
> Great work.

The code was thankfully pretty easy to read, even though it was written 2 years ago!

The OpenWRT wiki though was only kind of helpful, it's still a bit vague on some parts of UCI, so I had to look at the `uci.h` source-code to fully understand what was going on.

PR #496, opened 2023-03-07
Fix incorrect format specifier for an `int` in `process_get_bridges_cmd()`

aloisklink:
We're using `%s` to format an `int`, which is incorrect.

Fixes: b04ad88be46b67a335fd2f78ad34eec6e75a373f

---

Yep, I made this bug yesterday in PR https://github.com/nqminds/edgesec/pull/477 :facepalm:

In my defence, the `s` and `d` keys on my keyboard are quite close together.

### DIFF
--- a/src/supervisor/cmd_processor.c
+++ b/src/supervisor/cmd_processor.c
@@ -522,7 +522,7 @@ ssize_t process_get_bridges_cmd(int sock,
     {
       char *reallocated_reply_buf = os_realloc(reply_buf, total);
       if (reallocated_reply_buf == NULL) {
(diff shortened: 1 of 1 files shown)


mereacre (approved):
And I didn't see in my review :(

aloisklink:
> And I didn't see in my review `:(`

Once all of these PRs fixing the `log_*()` format bugs are merged, I've got a fix that makes the compiler throw an error if we ever make format bugs again!

After that, I'll have more complicated CHERIBuild work to do, so that should give you a break in reviewing PRs ;)

PR #501, opened 2023-03-07
Mark `log_*()` functions with `__attribute__((format))` for improved compiler warnings

aloisklink:
In `src/utils/log.h`, there is a currently unused macro (since https://github.com/nqminds/edgesec/commit/2a18535cb57f7059dba72eb50490bda587291fe8) called `PRINTF_FORMAT` that can be used to tell compilers that a given function passes arguments to `printf`.

https://github.com/nqminds/edgesec/blob/8cd691957f4b727f81d08a0e452caf22102de317/src/utils/log.h#L40-L44

This syntax is `__attribute__((format(printf, (a), (b))))`, and is supported by GCC and Clang. [^1]

See https://clang.llvm.org/docs/AttributeReference.html#format

This PR adds documentation to the macro, adds support for this macro to Clang, and then marks all the `log_*` functions in `src/utils/log.h` with the macro.

[^1]: `__attribute((format` is even supported by the [ARM compiler][1] and [IBM C/C++ compiler][2].

[1]: https://developer.arm.com/documentation/dui0375/g/Compiler-specific-Features/--attribute----format---function-attribute
[2]: https://www.ibm.com/docs/en/xl-c-and-cpp-aix/16.1?topic=attributes-format

The compiler errors that were generated led to me finding the following bugs:
  - #486
  - #487 
  - #488
  - #489 
  - #490 
  - #491 
  - #492 
  - #493 
  - #494 
  - #496 
  - #497 
  - #498 
  - #499 
  - #500

### DIFF
--- a/src/utils/log.h
+++ b/src/utils/log.h
@@ -85,12 +85,16 @@ void log_set_meta(bool enable);
 int log_open_file(char *path);
 void log_close_file(void);
 
+PRINTF_FORMAT(4, 5)
 void log_levels(uint8_t level, const char *file, uint32_t line,
                 const char *format, ...);
+PRINTF_FORMAT(4, 5)
 void log_errno_error(uint8_t level, const char *file, uint32_t line,
                      const char *format, ...);
+PRINTF_FORMAT(4, 5)
 void log_error_exit(uint8_t level, const char *file, uint32_t line,
                     const char *format, ...);
+PRINTF_FORMAT(4, 5)
 void log_error_exit_proc(uint8_t level, const char *file, uint32_t line,
                          const char *format, ...);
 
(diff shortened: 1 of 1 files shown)


aloisklink:
> Great catch.I caught a similar one in the BRSKI repo.

It's potentially worth adding a similar `__attribute__((format(vsnprintf` to `send_events_subscriber`:

https://github.com/nqminds/edgesec/blob/674b1feb9e4eb845405fccd53386466ed938136c/src/supervisor/subscriber_events.h#L41-L50

And a similar `__attribute((format(vfprintf` to `log_cmdline_error()`:

https://github.com/nqminds/edgesec/blob/674b1feb9e4eb845405fccd53386466ed938136c/src/edgesec.c#L95-L109

mereacre:
Yes, please add one to the even subscriber.

PR #504, opened 2023-03-13
docs(sockctl): fix typo in comment

aloisklink:
Considering I made this commit on a Monday, I'm going to assume it was because I was using the weak keyboard in the office, instead of a super nice mechanical keyboard from home. :smile:

I wonder if it's worth enabling something like [codespell](https://github.com/codespell-project/codespell). It easily integrates with `pre-commit, so it should be pretty easy to run. The only issue is fixing all the false positives.

### DIFF
--- a/src/utils/sockctl.c
+++ b/src/utils/sockctl.c
@@ -87,7 +87,7 @@ static int cleanup_tmp_domain_socket_path(const char *socket_path) {
   }
 
   // need to make a non-const copy of path since dirname() may change
-  // stirng contents
+  // string contents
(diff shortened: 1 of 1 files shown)


mereacre (approved):

mereacre:
If it is easily done with pre-commit, then we should integrate it.

aloisklink:
> If it is easily done with pre-commit, then we should integrate it.

Because there might be a bunch of false positives to fix, I've made an issue for it: https://github.com/nqminds/edgesec/issues/505

PR #506, opened 2023-03-13
Update `CodeCoverage.cmake` with upstream changes

aloisklink:
Update our `CodeCoverage.cmake` with the latest changes from upstream (aka https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake).

This is required for GCC compatibility, since the current `CodeCoverage.cmake` throws an error if we are using GCC, but not using GXX.

I did some investigating, and it looks like when our `CMakeModules/CodeCoverage.cmake` file was added in https://github.com/nqminds/edgesec/commit/acea20c1be2b5aae31c4a26b381a8b584a8462c7, this was equivalent to commit https://github.com/bilke/cmake-modules/commit/d98828f54f6974717798e63195cfbf08fe2daad0.

Then, I took the patches from https://github.com/bilke/cmake-modules/commit/d98828f54f6974717798e63195cfbf08fe2daad0 until https://github.com/bilke/cmake-modules/commit/877bab9dd1b17468c5d939cacaa2ad7ba99d1977 (current `master`) by running:

```console
user@pc:~/Documents/cmake-modules$ git format-patch --relative d98828f54f6974717798e63195cfbf08fe2daad0..877bab9dd1b17468c5d939cacaa2ad7ba99d1977 CodeCoverage.cmake
```

Then, I applied them to the edgesec repo using:

```console
# I had to fix a bunch of conflicts
user@pc:~/Documents/edgesec$ git am --directory CMakeModules/ ~/Documents/cmake-modules/000*.patch
```

I also squashed some of the patches from upstream together, to clean them up a bit.

---

@mereacre, should we use "Squash and merge" to merge this PR? It might simplify things, since I'm not 100% sure we need to see all of the individual changes in our Git history.

### DIFF
--- a/CMakeModules/CodeCoverage.cmake
+++ b/CMakeModules/CodeCoverage.cmake
@@ -83,6 +83,10 @@
 #     - Change gcovr output from -o <filename> for --xml <filename> and --html <filename> output respectively.
 #       This will allow for Multiple Output Formats at the same time by making use of GCOVR_ADDITIONAL_ARGS, e.g. GCOVR_ADDITIONAL_ARGS "--txt".
 #
+# 2022-09-28, Sebastian Mueller
+#     - fix append_coverage_compiler_flags_to_target to correctly add flags
+#     - replace "-fprofile-arcs -ftest-coverage" with "--coverage" (equivalent)
+#
 # USAGE:
 #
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Yes, please use squash and merge.

PR #509, opened 2023-03-13
Remove dependency on C++

aloisklink:
EDGESec only uses C code, so we don't need to get CMake to search for a C++ compiler [^1].

See https://askubuntu.com/a/1211755

OpenSSL still expects a CXX value, but it's not used for the OpenSSL build we compile, so we can just set it to an executable that does not exists. This is so we do start to use C++ in OpenSSL (e.g. due to a future OpenSSL update), we should instantly get an error.

[^1]: This occasionally causes issues with Clang, since `clang++` fails to compile a test program due to a missing `libstdc++`, because even though I have `g++-11` installed (and therefore `libstdc++-11-dev`), Clang sees that I have `gcc-12` installed, and incorrectly assumes that means I have `g++-12` (and GNU's `libstdc++-12-dev`) installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ project(EDGESEC
   VERSION 0.1.0
   HOMEPAGE_URL "https://github.com/nqminds/edgesec"
   DESCRIPTION "Secure router - reference implementation"
-  LANGUAGES C CXX
+  LANGUAGES C
 )
 # CMake proposal for semver https://gitlab.kitware.com/cmake/cmake/-/issues/23649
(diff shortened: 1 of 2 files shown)


aloisklink:
It looks like this is currently broken. I'm going to move this in "Draft" status, and we can re-open this later, if I get it working.

mereacre (approved):

aloisklink:
Working now that PR #512 and PR #510 have been merged!

This should also make our `cmake` configure time much faster, since CMake can skip all of their `C++` tests.

PR #514, opened 2023-03-13
test(runctl): fix radius_server_data heap overflow

aloisklink:
Fix a heap overflow error in `test_run_engine()`, when we weren't allocating enough bytes for a `struct radius_server_data`.

---

We may want to consider adding `__attribute__((alloc_size (1))` to the `os_zalloc()` function to tell the compiler that it returns the number of bytes given by the `size` parameter.

I think compilers/static analysers should hopefully warn us then about [CWE-131: Incorrect Calculation of Buffer Size](https://cwe.mitre.org/data/definitions/131.html).

For example, GCC has `-Wanalyzer-allocation-size` that would warn us.

### DIFF
--- a/tests/test_runctl.c
+++ b/tests/test_runctl.c
@@ -211,7 +211,7 @@ static void test_run_engine(void **state) {
 #endif
 #ifdef WITH_RADIUS_SERVICE
   struct radius_server_data *radius_srv =
-      os_zalloc(sizeof(struct radius_server_data *));
+      os_zalloc(sizeof(struct radius_server_data));
 #endif
 
   struct fwctx *fw_ctx = os_zalloc(sizeof(struct fwctx));


aloisklink:
> Merging [#514](https://codecov.io/gh/nqminds/edgesec/pull/514?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=nqminds) (aa87544) into [main](https://codecov.io/gh/nqminds/edgesec/commit/76dfb38c3630e94808d44b3e934dc0c898ad2306?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=nqminds) (76dfb38) will **increase** coverage by `0.98%`.

Looks like code-coverage has greatly increased too (a 1% increase is a lot!! We've only increased by 2.6% in the last 6 months!)

I'm guessing that the this heap overflow bug caused a bunch of `test_runctl.c` tests to not run correctly!

mereacre (approved):

PR #518, opened 2023-03-13
test(edgesec): fix `test_edgesec()` config memory leak

aloisklink:
Some of the values in `config` were allocated on the heap by `load_app_config()`:

https://github.com/nqminds/edgesec/blob/cbd6701393d7cd5ff3b134fb46056f206e2704b9/tests/test_edgesec.c#L350

Because of that, we must call `free_app_config()` to deallocate this variables.

---

Should we document this in the `load_app_config()` documentation?

https://github.com/nqminds/edgesec/blob/429d6179ec00553fe1959c105fb6dcb9a3ea4018/src/config.h#L86


mereacre:
Yes, definitely.

mereacre (approved):

PR #526, opened 2023-03-14
Fix buffer overflow with ifname in `iface_mapper`

aloisklink:
Fix buffer overflows using the `ifname` parameter in the following functions in `src/utils/iface_mapper.c`:
  - `get_if_mapper()`
  - `put_if_mapper()`

Currently, we're using `os_memcpy(ifname, s->value, IF_NAMESIZE);` to copy the string from `ifname`. This is fine if `ifname` is at least IF_NAMESIZE bytes, but it's instead a NUL-terminated string that may be less than IF_NAMESIZE bytes long. In that case, we must only read up to `strlen(ifname)` bytes.

This can be fixed by replacing the `memcpy()` calls with `os_strlcpy()` (edgesec's implementation of [BSD strlcpy()](https://man.freebsd.org/cgi/man.cgi?query=strlcpy&sektion=3) function).

### DIFF
--- a/src/utils/iface_mapper.c
+++ b/src/utils/iface_mapper.c
@@ -43,7 +43,7 @@ int get_if_mapper(hmap_if_conn **hmap, in_addr_t subnet, char *ifname) {
             s); /* id already in the hash? */
 
   if (s != NULL) {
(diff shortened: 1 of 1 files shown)


mereacre (approved):
Good catch.

PR #532, opened 2023-03-17
Cleanup `ctx.crypt_ctx` correctly by calling `free_crypt_service()`

aloisklink:
The `test_get_mac_conn_cmd` test in `test_supervisor.c` was not correctly cleaning up the `ctx.crypt_ctx` object, when we were compiling tests with `WITH_CRYPTO_SERVICE=ON`.

We should be using `free_crypt_service()` to avoid a memory leak.

Fixes: https://github.com/nqminds/edgesec/commit/0913b22966baa4f0ef7c1518fc636e2a534b565d

---

In order to prevent this from happening again, I've added documentation to the `load_crypt_service()` function to say that `free_crypt_service()` should be called on the result. If we're compiling with GCC >= 11, I've also added the `__attribute__((malloc(free_crypt_service, 1)))` compiler attribute to `load_crypt_service()`, so that the compiler warns us if we use the wrong deallocator.

### DIFF
--- a/src/crypt/CMakeLists.txt
+++ b/src/crypt/CMakeLists.txt
@@ -18,4 +18,4 @@ else ()
 endif ()
 
 add_library(crypt_service crypt_service.c)
-target_link_libraries(crypt_service PUBLIC LibUTHash::LibUTHash SQLite::SQLite3 PRIVATE generic_hsm_driver base64 sqlite_crypt_writer cryptou log os)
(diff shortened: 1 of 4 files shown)


mereacre (approved):

aloisklink:
Hmmmm, looks like the improved compiler warnings seem to be catching a bunch of other potential bugs. I'll convert this PR to a draft, and I guess this is something to fix later when we have more free time.

aloisklink:
Fixed!

The issue was that I made a typo in the following bit of code:

```c
#define __must_free_crypt_service                                              \
  __attribute__((malloc(free_crypt_service, 1))) __must_check
#else
// I accidentally did #define __must_free here!
#define __must_free_crypt_service __must_check
#endif /* __GNUC__ >= 11 */
```

PR #541, opened 2023-03-27
build(cmake): enable UndefinedBehaviorSanitizer

aloisklink:
Enable [UndefinedBehaviorSanitizer (UBSan)][1] on Linux presets by default.

> UndefinedBehaviorSanitizer (UBSan) is a fast undefined behavior detector. UBSan modifies the program at compile-time to catch various kinds of undefined behavior during program execution, for example:
> 
> - Array subscript out of bounds, where the bounds can be statically determined
> - Bitwise shifts that are out of bounds for their data type
> - Dereferencing misaligned or null pointers
> - Signed integer overflow
> - Conversion to, from, or between floating-point types which would overflow the destination

The OpenWRT SDK presets have it disabled, as old versions of GCC have poor support for running UBSanitizer on non-x86 platforms.

[1]: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,6 +297,29 @@ if (NOT BUILD_ONLY_DOCS)
     )
   endif()
 
+  if (SANITIZE_UNDEFINED_BEHAVIOR)
+  include(CheckCCompilerFlag)
+
+  set(CMAKE_REQUIRED_LINK_OPTIONS "-fsanitize=undefined")
+  check_c_compiler_flag("-fsanitize=undefined" C_COMPILER_SUPPORTS_UBSAN)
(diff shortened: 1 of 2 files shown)


aloisklink:
> I added the same lines to my BRSKI code and doesn't seem to find the asn libray.

Do you mean it couldn't find the `UBSan` library or the `ASan` library? It might not be supported for your cross-compiling configuration, or maybe your configuration (C++) means that ASan and UBSan can't run at the same time?

I know that ThreadSanitizer and MemorySanitizer can't run at the same time as AddressSanitizer, so maybe that could be the issue?

PR #544, opened 2023-03-27
refactor(dnsmasq): make `signal_dhcp_process()` const

aloisklink:
Make the `dhcp_bin_path` parameter to `signal_dhcp_process()` `const`.

Because we call `basename()` on this value, which may change the input string, we need to first make a temporary copy of the string into a buffer.


mereacre (approved):

aloisklink:
Awesome, it looks like this fixed a failing test in FreeBSD :)

Apparently running `basename("this-is-a-string-constant");` on Linux is fine, but on FreeBSD it causes a segmentation fault!

